PR #7, opened 2020-02-19
3540: API, fxcop implementation

RuthBerry96:


Backend/src/NHSD.BuyingCatalogue.Identity.Api/Infrastructure/LogHelper.cs
```diff
@@ -9,7 +9,9 @@ public static class LogHelper
     {
         public static void EnrichFromRequest(IDiagnosticContext diagnosticContext, HttpContext httpContext)
         {
-            var request = httpContext.Request;
+            var request = httpContext.ThrowIfNull().Request;
+
+            diagnosticContext = diagnosticContext.ThrowIfNull();
```
sam-roberts-bjss: Can you review this assignment? I'm wondering why it is assigned to itself.

Backend/src/NHSD.BuyingCatalogue.Identity.Api/Startup.cs
```diff
@@ -16,6 +16,7 @@ public Startup(IWebHostEnvironment environment)
             Environment = environment;
         }
 
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Performance", "CA1822:Mark members as static", Justification = "ASP.net needs this to not be static")]
```
sam-roberts-bjss: To be consistent with all the other projects, could this be moved to the top of the file. Taken from the buying Catalogue API

--------------

    [SuppressMessage("Design", "CA1822", Justification = "ASP.Net needs this to not be static")]
    public sealed class Startup

--------------

PR #12, opened 2020-02-24
3540: Identity, ReadMe

RuthBerry96:


README.md
```diff
@@ -1 +1,65 @@
...
+### Running
+To start up the API, run the command from the root directory of the repository.
+
+```bash
+docker-compose up -d --build
+```
+
+This will start the API in a docker container.
+
+You can verifity that the API has launched correctly by navigating to the following url via any web browser, using the following test data. This should successfully log in a user and bring back a return url.
```
sam-roberts-bjss: verifity? Check spelling here please

README.md
```diff
@@ -1 +1,65 @@
...
+
+### Architecture overview
+This application uses **.NET core** to provide an API capable of runnning Linux or Windows.
+
+### Overview of the application code
+This repository uses **.NET Core** and **Docker**.
+
+It contains three main endpoints
+
+- account/login/returnUrl
```
sam-roberts-bjss: Please review these endpoints.

The login endpoint will be something like 'account/login?returnUrl=<Return URL Here>'

PR #14, opened 2020-02-24
5166: Add node to docker image

sam-roberts-bjss:


.dockerignore
```diff
@@ -10,4 +10,6 @@ Dockerfile
 
 bin
 obj
-*.*proj.user
\ No newline at end of file
+*.*proj.user
+
+src\NHSD.BuyingCatalogue.Identity.Api\node_modules
```
stuart-bass-bjss: Can we make this **/node_modules please?

PR #16, opened 2020-02-24
3540: Adding Docker for Sample Client & making sample resource work

CallumBradburyBJSS:


src/NHSD.BuyingCatalogue.Identity.Api/Startup.cs
```diff
@@ -34,7 +35,7 @@ public void ConfigureServices(IServiceCollection services)
 
             services.AddIdentityServer(options =>
             {
-                options.IssuerUri = "null";
+                options.IssuerUri = "my_auth";
```
stuart-bass-bjss: Can we please move this into config?

PR #22, opened 2020-02-27
Dockerize the new identity database

bjss-chris:
AB#5267


docker-compose.yml
```diff
@@ -1,5 +1,23 @@
...
+    build: 
+      context: src/NHSD.BuyingCatalogue.Identity.UserDatabase
+    container_name: identity_api_user_db
+    volumes:
+      - nhsd_identity_user_db_volume:/var/opt/mssql/
+    networks:
+      - nhsd_bcnetwork
+    environment:
+      - ACCEPT_EULA=Y
+      - SA_PASSWORD=MyTestPassword123!
```
CallumBradburyBJSS: My initial impulse is that this being different to the value in the sqlproj is going to make me sad somewhere down the line

bjss-chris: That's the `sa` password. The test password for the `NHSD` user is the same as in the sqlproj.

PR #23, opened 2020-02-27
3540: Updated Selenium tests to launch locally when debugger is attached

CallumBradburyBJSS:


tests/NHSD.BuyingCatalogue.Identity.Api.IntegrationTests/Utils/SeleniumContext.cs
```diff
@@ -14,8 +15,16 @@ public sealed class SeleniumContext : IDisposable
...
-            WebDriver = new RemoteWebDriver(new Uri("http://localhost:4444/wd/hub"), options);
+            options.AddArguments("window-size=1920,1080", "no-sandbox", "disable-dev-shm-usage");
+            if (!Debugger.IsAttached)
+            {
+                options.AddArgument("headless");
+                WebDriver = new RemoteWebDriver(new Uri("http://localhost:4444/wd/hub"), options);
+            }
+            else
+            {
+                WebDriver = new ChromeDriver(options);
```
PhysicsJon: As a suggestion, you could add "auto-open-devtools-for-tabs" flags when running locally as this will make debugging even easier (will start recording network requests immediately)

PR #25, opened 2020-02-28
3540: Updated Login Integration Tests to run via MVC Sample Client

CallumBradburyBJSS:


samples/NHSD.BuyingCatalogue.Identity.Api.SampleMvcClient/Controllers/HomeController.cs
```diff
@@ -44,10 +44,9 @@ public async Task<IActionResult> Privacy()
             }
             else
             {
-                var content = await response.Content.ReadAsStringAsync();
-                Console.WriteLine(JArray.Parse(content));
-                ViewBag.Response = JArray.Parse(content);
+                ViewBag.Response = "Authorized With Sample Resource";
             }
+            Console.WriteLine(ViewBag.Response);
```
sam-roberts-bjss: Could you review this line please? I think we don't need this extra logging here unless it is used for something I am missing.

tests/NHSD.BuyingCatalogue.Identity.Api.IntegrationTests/Features/Login with a valid client 3540.feature
```diff
@@ -10,6 +10,22 @@ Background:
...
+
+@3540
+Scenario: 2. Logging in with an existing client with invalid username
+    When the user navigates to a restricted web page
+    Then the user is redirected to identity server page account/login
+    When a login request is made with username Invalid and password Pass123$
+    Then the user is redirected to identity server page account/login
+
+@3540
+Scenario: 3. Logging in with an existing client with invalid username
```
sam-roberts-bjss: Should this read invalid password rather than invalid username?

tests/NHSD.BuyingCatalogue.Identity.Api.IntegrationTests/Features/Login with a valid client 3540.feature
```diff
@@ -10,6 +10,22 @@ Background:
 
 @3540
 Scenario: 1. Logging in with an existing client with valid credentials
-    When the user navigates to the login page with return url http://localhost:8072
+    When the user navigates to a restricted web page
+    Then the user is redirected to identity server page account/login
```
sam-roberts-bjss: I would be tempted to reword this to:
Then the user is redirected to the login page. 

My thinking here is does the user really need to know its an Identity server? In future we may need to distinguish between the Buying Catagloue login page and the NHS Email login page.

tests/NHSD.BuyingCatalogue.Identity.Api.IntegrationTests/Steps/LoginSteps.cs
```diff
@@ -8,18 +11,25 @@ namespace NHSD.BuyingCatalogue.Identity.Api.IntegrationTests.Steps
     [Binding]
     internal sealed class LoginSteps
     {
+        private readonly Dictionary<string, string> HostUrls = new Dictionary<string, string>
+        {
+            {"identity server", "http://host.docker.internal:8070/"},
```
sam-roberts-bjss: It would be good to get the base addresses into configuration. This way we only need to configure it once. It can be a separate task if you see fit.

tests/NHSD.BuyingCatalogue.Identity.Api.IntegrationTests/Steps/LoginSteps.cs
```diff
@@ -30,10 +40,16 @@ public void WhenALoginRequestIsMade(string username, string password)
...
+        public void ThenTheUserIsRedirectedTo(string target, string url)
+        {
+            _seleniumContext.WebWaiter.Until(x => x.Url.StartsWith($"{HostUrls[target]}{url}",StringComparison.OrdinalIgnoreCase));
+        }
+
+        [Then(@"the page contains paragraph with text (.*)")]
+        public void ThenThePageVerifiesItCouldTalkToTheSampleResource(string text)
         {
-            _seleniumContext.WebWaiter.Until(x => x.Url == url);
+            _seleniumContext.WebDriver.FindElement(By.Id("sampleResourceResult")).Text.Should().Be(text);
```
sam-roberts-bjss: We had a quick discussion on this. I thought we were going to try keep this a bit more generic? By making it generic to find a particular tag or element we will introduce a set of common steps that other tests will be able to use.

samples/NHSD.BuyingCatalogue.Identity.Api.SampleMvcClient/Controllers/HomeController.cs
CallumBradburyBJSS: Yeah we probably don't need it, I was just tidying up the pre-existing Console.WriteLine's to be simpler, instead of it appearing inside both the If and the else. Can strip it out for sure!

tests/NHSD.BuyingCatalogue.Identity.Api.IntegrationTests/Steps/LoginSteps.cs
CallumBradburyBJSS: Sure, we can do that

tests/NHSD.BuyingCatalogue.Identity.Api.IntegrationTests/Features/Login with a valid client 3540.feature
CallumBradburyBJSS: hmm, I initially added the identity server vs client parameter to the redirected page, so that we had one generic step for redirections that could deal with the differing ports for identity vs client. I see what you mean though, it being an identity server is somewhat irrelevant. Maybe I could change it from identity server to... authentication page, or something like that, as the parameter passed in?

sam-roberts-bjss: Yeah I don't think the user needs to know whether it is a client or a server. All they really know is a web page. Yeah definitely rename it. I feel like login page would be better, authentication sounds more technical to a general user.

PR #26, opened 2020-03-01
3568 Add logout

sam-roberts-bjss:


src/NHSD.BuyingCatalogue.Identity.Api/Services/LogoutService.cs
```diff
@@ -0,0 +1,45 @@
...
+
+            HttpContext httpContext = _httpContextAccessor.HttpContext;
+
+            await httpContext.SignOutAsync();
+            await httpContext.SignOutAsync(IdentityConstants.ApplicationScheme);
+
+            httpContext.User = new ClaimsPrincipal(new ClaimsIdentity());
+        }
+    }
+}
```
bjss-chris: Do we want to raise a `UserLogoutSuccessEvent` here to tie in with the login events? 

(Would obviously require an additional dependency on `IEventService`.)

src/NHSD.BuyingCatalogue.Identity.Api/Startup.cs
```diff
@@ -59,15 +61,14 @@ public void ConfigureServices(IServiceCollection services)
...
         }
 
         public void Configure(IApplicationBuilder app)
         {
             app.UseSerilogRequestLogging(opts =>
             {
-                opts.EnrichDiagnosticContext = LogHelper.EnrichFromRequest;
-                opts.GetLevel = LogHelper.ExcludeHealthChecks;
+                opts.EnrichDiagnosticContext = Infrastructure.LogHelper.EnrichFromRequest;
+                opts.GetLevel = Infrastructure.LogHelper.ExcludeHealthChecks;
```
bjss-chris: Why remove the `using` alias?

I think renaming/refactoring `LogHelper` would probably be a better long-term option (the name is vague and `ExcludeHealthChecks` is also badly named in my opinion – probably a tech debt ticket though.

tests/NHSD.BuyingCatalogue.Identity.Api.UnitTests/Builders/AccountControllerBuilder.cs
```diff
@@ -0,0 +1,55 @@
+﻿using IdentityServer4.Services;
+using IdentityServer4.Stores;
+using Microsoft.AspNetCore.Identity;
```
bjss-chris: Is this required?

tests/NHSD.BuyingCatalogue.Identity.Api.UnitTests/Builders/AccountControllerBuilder.cs
```diff
@@ -0,0 +1,55 @@
...
+using Moq;
+using NHSD.BuyingCatalogue.Identity.Api.Controllers;
+using NHSD.BuyingCatalogue.Identity.Api.Models;
```
bjss-chris: Is this required?

tests/NHSD.BuyingCatalogue.Identity.Api.UnitTests/Controllers/AccountControllerTests.cs
```diff
@@ -0,0 +1,112 @@
...
+
+            using var sut = new AccountControllerBuilder()
+                .WithLogoutService(logoutServiceMock.Object)
+                .Build();
+
+            await sut.Logout(new LogoutViewModel());
+
+            logoutServiceMock.Verify(x => x.GetPostLogoutRedirectUri(
+                It.Is<string>(actual => actual == null)), Times.Once);
+        }
```
bjss-chris: Could you explain your thoughts on this test? Unless I've missed something won't `GetPostLogoutRedirectUri` only ever be called once regardless of the state `logoutViewModel` (except the `null` case)?

src/NHSD.BuyingCatalogue.Identity.Api/appsettings.json
```diff
@@ -9,8 +9,8 @@
       "requireClientSecret": true,
       "secret": "SampleClientSecret",
       "requireConsent": false,
-      "redirectUrls": [ "http://localhost:49176/signin-oidc" ],
-      "postLogoutRedirectUrls": [],
+      "redirectUrls": [ "http://localhost:8072/signin-oidc" ],
```
bjss-chris: Why 8072? It looks like `launchSettings.json` still defaults to 49176 for the sample MVC client.

It might be worth looking at the default `launchSettings.json` port settings across the sample client, resource, and IDS projects as I have to change them every time I want to debug locally – unless Stuart is planning to do so as part of his dev environment changes...

bjss-chris: Sorry, just spotted that it's the port configured in `docker-compose.yml`.

(Although that doesn't resolve the port issue when not using the dockerized version.)

samples/NHSD.BuyingCatalogue.Identity.Api.SampleMvcClient/Controllers/HomeController.cs
```diff
@@ -56,5 +61,18 @@ public IActionResult Error()
...
+        {
+            _logger.LogInformation("Start Logging out user from client");
+
+            await HttpContext.SignOutAsync(CookieAuthenticationDefaults.AuthenticationScheme);
+            await HttpContext.SignOutAsync(OpenIdConnectDefaults.AuthenticationScheme);
+
+            _logger.LogInformation("Stop Logging out user from client");
+
+            var homeUrl = Url.Action(nameof(Index), "Home");
+            return new SignOutResult(OpenIdConnectDefaults.AuthenticationScheme, new AuthenticationProperties { RedirectUri = homeUrl });
```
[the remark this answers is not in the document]
sam-roberts-bjss: Yeah good shout. Looks like it changes the order of parameters but not a problem.

samples/NHSD.BuyingCatalogue.Identity.Api.SampleMvcClient/Startup.cs
```diff
@@ -22,27 +23,30 @@ public void ConfigureServices(IServiceCollection services)
...
+            })
+            .AddCookie(CookieAuthenticationDefaults.AuthenticationScheme)
+            .AddOpenIdConnect(OpenIdConnectDefaults.AuthenticationScheme, options =>
+            {
+                options.SignInScheme = CookieAuthenticationDefaults.AuthenticationScheme;
+                options.Authority = authority;
+                options.SignedOutRedirectUri = signedOutRedirectUri;
+                options.ClientId = clientId;
+                options.ClientSecret = clientSecret;
+                options.ResponseType = "code";
```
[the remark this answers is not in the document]
sam-roberts-bjss: 👍

samples/NHSD.BuyingCatalogue.Identity.Api.SampleMvcClient/Views/Shared/_Layout.cshtml
```diff
@@ -1,4 +1,6 @@
-﻿<!DOCTYPE html>
+﻿@using Microsoft.AspNetCore.Identity
```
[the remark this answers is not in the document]
sam-roberts-bjss: Nope, good spot, not required.

src/NHSD.BuyingCatalogue.Identity.Api/Services/LogoutService.cs
sam-roberts-bjss: I'm still not sure about the events, whether or not they are adding anything. But we should be consistent so will add it.

src/NHSD.BuyingCatalogue.Identity.Api/Startup.cs
sam-roberts-bjss: I find that style of using confusing (one to add to the next coding style meeting). The name was clashing so had to be explicit.

I agree the name of the LogHelper & ExcludeHealthChecks should be changed.

tests/NHSD.BuyingCatalogue.Identity.Api.UnitTests/Builders/AccountControllerBuilder.cs
sam-roberts-bjss: Nope, cleaned them up 👍

tests/NHSD.BuyingCatalogue.Identity.Api.UnitTests/Controllers/AccountControllerTests.cs
sam-roberts-bjss: The idea here is to test dependencies used within the method. So when I call that method I want to ensure that my dependency is getting called only once & how I expect it. I didn't want to do it in every test because if the method changes every test would fail. At least this way there is the potential for only this one that would fail.

This was a new concept for me in unit tests. The previous tech lead was keen on this. In the past I have only ever tested input output. But one could class the dependency as input I guess.

src/NHSD.BuyingCatalogue.Identity.Api/appsettings.json
sam-roberts-bjss: This was used during manual testing. Stuart said he would be changing these shortly. So I left it. I'll change them back to match the launch settings of the MVC client e.g. 49176

src/NHSD.BuyingCatalogue.Identity.Api/Controllers/AccountController.cs
```diff
@@ -99,6 +105,28 @@ public async Task<IActionResult> Login(LoginViewModel viewModel)
             throw new Exception("Invalid return URL");
         }
 
+        [HttpGet]
+        public async Task<IActionResult> Logout(string logoutId)
```
bjss-chris: Do we want to validate that `logoutId` is not null, empty or white space here? (I believe `GetLogoutRequest` will return `null` in that case and `Redirect` will throw if passed a null or empty string.)

Also, do we definitely want/need to provide GET given that a GET shouldn't have side effects?

src/NHSD.BuyingCatalogue.Identity.Api/Controllers/AccountController.cs
```diff
@@ -99,6 +105,28 @@ public async Task<IActionResult> Login(LoginViewModel viewModel)
...
+        {
+            return await Logout(new LogoutViewModel
+            {
+                LogoutId = logoutId
+            });
+        }
+
+        [HttpPost]
+        [ValidateAntiForgeryToken]
```
bjss-chris: Could this cause us problems when we integrate the Node.js front-end? (Although we obviously need to mitigate the CSRF threat.)

src/NHSD.BuyingCatalogue.Identity.Api/Controllers/AccountController.cs
```diff
@@ -99,6 +105,28 @@ public async Task<IActionResult> Login(LoginViewModel viewModel)
...
+        public async Task<IActionResult> Logout(LogoutViewModel logoutViewModel)
```
bjss-chris: Should we have an `Authorize` attribute here? While redirecting to login just to logout again may be pointless is it worth considering to prevent hitting logout unless a user is already logged in?

src/NHSD.BuyingCatalogue.Identity.Api/Controllers/AccountController.cs
```diff
@@ -99,6 +105,28 @@ public async Task<IActionResult> Login(LoginViewModel viewModel)
...
+        {
+            logoutViewModel.ThrowIfNull(nameof(logoutViewModel));
```
bjss-chris: Is a null or empty `logoutViewModel.LogoutId` a valid scenario? As we'll end up with a null `logoutRequest` when `logoutViewModel.LogoutId` is null or empty and  `Redirect` will throw in that case is it worth adding `Required` attribute to the `LogoutViewModel.LogoutId`? We can then check whether `ModelState.IsValid`.

src/NHSD.BuyingCatalogue.Identity.Api/Controllers/AccountController.cs
```diff
@@ -99,6 +105,28 @@ public async Task<IActionResult> Login(LoginViewModel viewModel)
...
+
+            LogoutRequest logoutRequest = await _logoutService.GetLogoutRequest(logoutViewModel.LogoutId);
+
+            await _logoutService.SignOutAsync(logoutRequest);
+
+            return Redirect(logoutRequest?.PostLogoutRedirectUri);
```
bjss-chris: As `Redirect` will throw if the URL passed to it is null or empty is it worth a change of approach here? The other option is to allow `Redirect` to throw and the user should be directed to the error page (once that change is merged in).

src/NHSD.BuyingCatalogue.Identity.Api/Services/LogoutService.cs
```diff
@@ -0,0 +1,49 @@
...
+        public async Task SignOutAsync(LogoutRequest logoutRequest)
+        {
+            HttpContext httpContext = _httpContextAccessor.HttpContext;
+
+            await httpContext.SignOutAsync();
+            await httpContext.SignOutAsync(IdentityConstants.ApplicationScheme);
+
+            httpContext.User = new ClaimsPrincipal(new ClaimsIdentity());
+
+            if (logoutRequest is object)
```
bjss-chris: Would `logoutRequest != null` not be slightly clearer here? (Purely from intent/readability point-of-view.)

src/NHSD.BuyingCatalogue.Identity.Api/ViewModels/Account/LogoutViewModel.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace NHSD.BuyingCatalogue.Identity.Api.ViewModels.Account
+{
+    public sealed class LogoutViewModel
+    {
+        public string LogoutId { get; set; }
```
bjss-chris: Worth adding a `Required` attribute here as above?

samples/NHSD.BuyingCatalogue.Identity.Api.SampleMvcClient/Controllers/HomeController.cs
```diff
@@ -4,9 +4,12 @@
 using System.Net.Http.Headers;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Authentication;
+using Microsoft.AspNetCore.Authentication.Cookies;
+using Microsoft.AspNetCore.Authentication.OpenIdConnect;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Logging;
 using Newtonsoft.Json.Linq;
```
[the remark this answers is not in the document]
sam-roberts-bjss: 👍

src/NHSD.BuyingCatalogue.Identity.Api/Controllers/AccountController.cs
sam-roberts-bjss: Yeah I will guard against the logoutId for null or white space. The GET is required as the logout would not work without it. Usually you would want to add a log out confirmation page here. Currently for this application we don't need it.

sam-roberts-bjss: There is the possibility to call  ```_interaction.CreateLogoutContextAsync(); ``` to create a logout Id if one is not supplied. Unfortunately this can still return null if there is not user. I have seen this used when using an external login / logout approach.

src/NHSD.BuyingCatalogue.Identity.Api/Controllers/AccountController.cs
sam-roberts-bjss: At the moment no client should be calling this endpoint. So I will remove it.

src/NHSD.BuyingCatalogue.Identity.Api/Controllers/AccountController.cs
sam-roberts-bjss: The few samples I have seen on this have allowed anonymous access. I see this as attempting to delete a user that has already been deleted. Don't really need to tell the user that. They got what they wanted.

src/NHSD.BuyingCatalogue.Identity.Api/Controllers/AccountController.cs
sam-roberts-bjss: I have removed this method & moved the logic into the GET logout method. Do you think the required tag would still apply?

src/NHSD.BuyingCatalogue.Identity.Api/Controllers/AccountController.cs
sam-roberts-bjss: I think the Redirect will throw an exception. I'm not sure what to do if the PostLogoutRedirectUri is null other than redirecting to the route '~/' but either way is not great for the user. I suppose the error page would give some feedback as least.

src/NHSD.BuyingCatalogue.Identity.Api/Controllers/AccountController.cs
bjss-chris: I wouldn't think so as there won't be any model binding happening. Is `LogoutViewModel` also still a view model?

src/NHSD.BuyingCatalogue.Identity.Api/Services/LogoutService.cs
sam-roberts-bjss: I thought this was the new standard for checking not null?

src/NHSD.BuyingCatalogue.Identity.Api/ViewModels/Account/LogoutViewModel.cs
sam-roberts-bjss: Deleted the class, no longer required.

src/NHSD.BuyingCatalogue.Identity.Api/Services/LogoutService.cs
bjss-chris: Interesting – do you have any references?

sam-roberts-bjss: I had a quick look for a reference but can't find the original (how convenient I know!), did find

https://github.com/dotnet/roslyn/issues/39253

This quote is from the above rabbit hole in answer to "What makes that the best null check in c#? I tend to use "x != null?":

``` 
"That is probably a null check but not guaranteed. If "x" has a custom equality operator then you get what the implementation of that gets. If "x" is a struct then there is a chance it's just evaluated as "false". Basically whether or not it's a null check is subject to context"
```

src/NHSD.BuyingCatalogue.Identity.Api/Controllers/AccountController.cs

src/NHSD.BuyingCatalogue.Identity.Api/Controllers/AccountController.cs
bjss-chris: Probably worth a follow-up task...

tests/NHSD.BuyingCatalogue.Identity.Api.UnitTests/Controllers/AccountControllerTests.cs
```diff
@@ -0,0 +1,106 @@
...
+            {
+                using var sut = new AccountControllerBuilder()
+                    .Build();
+
+                await sut.Logout(null);
+            });
+        }
+
+        [Test]
+        public void Logout_WhenLogoutIdIsEmpty_ShouldThrowArgumentNullException()
```
bjss-chris: Could this test and `Logout_WhenLogoutIdIsNull_ShouldThrowArgumentNullException` be combined into single test using test cases? If so, is it worth adding a test case for white space too?

PR #27, opened 2020-03-02
3540 add error view

bjss-chris:
Adds an error view for handling potentially malicious return URLs.

AB#5273


src/NHSD.BuyingCatalogue.Identity.Api/Controllers/AccountController.cs
```diff
@@ -55,26 +58,20 @@ public async Task<IActionResult> Login(LoginViewModel viewModel)
         {
             viewModel.ThrowIfNull(nameof(viewModel));
 
-            var returnUrl = viewModel.ReturnUrl.ToString();
-            var context = await _interaction.GetAuthorizationContextAsync(returnUrl);
+            var returnUrl = viewModel.ReturnUrl?.ToString();
```
bjss-chris: There may be a question as to whether we really need to handle the `ReturnUrl == null` case here because the anti-forgery validation essentially limits the invocation of this method to the MVC front-end.

As we set the return URL to the root (in the `GET` method) when the `returnUrl` is `null` we may not need the null-conditional operator.

src/NHSD.BuyingCatalogue.Identity.Api/Controllers/AccountController.cs
```diff
@@ -92,20 +89,37 @@ public async Task<IActionResult> Login(LoginViewModel viewModel)
             if (Url.IsLocalUrl(returnUrl))
                 return Redirect(returnUrl);
 
-            if (string.IsNullOrEmpty(returnUrl))
+            if (string.IsNullOrWhiteSpace(returnUrl))
```
bjss-chris: Following on from the comments above about the likelihood of `ReturnUrl == null` we may not need this check either.

src/NHSD.BuyingCatalogue.Identity.Api/Controllers/AccountController.cs
```diff
@@ -92,20 +89,37 @@ public async Task<IActionResult> Login(LoginViewModel viewModel)
...
                 return Redirect("~/");
 
-            // User might have clicked on a malicious link - should be logged
-            throw new Exception("Invalid return URL");
+            // Return URL could be malicious
+            throw new InvalidReturnUrlException();
         }
 
-        [HttpGet]
```
bjss-chris: The [Microsoft docs](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/error-handling?view=aspnetcore-3.1#exception-handler-page) state that error handler action methods should not be marked with HTTP method attributes as they can prevent some requests from reaching the method – this appears to be the case here.

src/NHSD.BuyingCatalogue.Identity.Api/Controllers/AccountController.cs
```diff
@@ -92,20 +89,37 @@ public async Task<IActionResult> Login(LoginViewModel viewModel)
...
         public async Task<IActionResult> Error(string errorId)
```
bjss-chris: At this stage of development, `errorId` will always be `null` so we do we actually require it?

This also affects the design of the error view model – please see the comments below about that.

src/NHSD.BuyingCatalogue.Identity.Api/Controllers/AccountController.cs
```diff
@@ -92,20 +89,37 @@ public async Task<IActionResult> Login(LoginViewModel viewModel)
...
         {
             // retrieve error details from identityserver
             var message = await _interaction.GetErrorContextAsync(errorId);
 
-            return Ok(message);
+            return View(new ErrorViewModel { Message = message?.ErrorDescription });
```
bjss-chris: As `errorId` will only ever be `null` at this stage of development, the same applies to `message`. Would we ever realistically want to display this error information?

src/NHSD.BuyingCatalogue.Identity.Api/ViewModels/ErrorViewModel.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace NHSD.BuyingCatalogue.Identity.Api.ViewModels
```
bjss-chris: This view model may not even be required at this point, depending on what we decide based on the comments above.

src/NHSD.BuyingCatalogue.Identity.Api/Views/Shared/Error.cshtml
```diff
@@ -0,0 +1,22 @@
...
+
+                    @if (Model.ShowMessage)
+                    {
+                        <p>@Model.Message</p>
+                    }
+                </div>
+            </div>
+        </div>
+    </main>
+</div>
```
bjss-chris: Additional work will still be required to make sure that the styling and content of the error page is consistent with the Node.js version.

src/NHSD.BuyingCatalogue.Identity.Api/Controllers/AccountController.cs
CallumBradburyBJSS: We'd probably want to provide the error message in development builds? Maybe?

src/NHSD.BuyingCatalogue.Identity.Api/Controllers/AccountController.cs
```diff
@@ -1,13 +1,16 @@
...
+using IdentityServer4.Models;
 using IdentityServer4.Services;
 using IdentityServer4.Stores;
 using Microsoft.AspNetCore.Identity;
 using Microsoft.AspNetCore.Mvc;
+using NHSD.BuyingCatalogue.Identity.Api.Exceptions;
 using NHSD.BuyingCatalogue.Identity.Api.Infrastructure;
 using NHSD.BuyingCatalogue.Identity.Api.Models;
 using NHSD.BuyingCatalogue.Identity.Api.ViewModels;
+using SignInResult = Microsoft.AspNetCore.Identity.SignInResult;
```
sam-roberts-bjss: I think this can be removed if you change the variable to use var instead of the actual type?

src/NHSD.BuyingCatalogue.Identity.Api/Controllers/AccountController.cs
sam-roberts-bjss: Yeah I agree I don't think we need to worry here about the Null reference. I don't mind belt & braces though just to be sure. Most of the samples I have seen don't seem to worry about it.

src/NHSD.BuyingCatalogue.Identity.Api/Controllers/AccountController.cs
```diff
@@ -55,26 +58,20 @@ public async Task<IActionResult> Login(LoginViewModel viewModel)
...
             if (!ModelState.IsValid)
                 return View(NewLoginViewModel());
 
-            var result = await _signInManager.PasswordSignInAsync(viewModel.Username, viewModel.Password, false, true);
-            if (!result.Succeeded)
-            {
-                await _eventService.RaiseAsync(new UserLoginFailureEvent(viewModel.Username, "invalid credentials", clientId: context?.ClientId));
-                ModelState.AddModelError(string.Empty, "Invalid username or password");
-
+            var signedIn = await SignIn(viewModel, context);
```
sam-roberts-bjss: Could we add Async to the SignIn method?

src/NHSD.BuyingCatalogue.Identity.Api/Controllers/AccountController.cs
sam-roberts-bjss: Yeah I agree.

src/NHSD.BuyingCatalogue.Identity.Api/Controllers/AccountController.cs
```diff
@@ -92,20 +89,37 @@ public async Task<IActionResult> Login(LoginViewModel viewModel)
...
                 return Redirect(returnUrl);
 
-            if (string.IsNullOrEmpty(returnUrl))
+            if (string.IsNullOrWhiteSpace(returnUrl))
                 return Redirect("~/");
 
-            // User might have clicked on a malicious link - should be logged
-            throw new Exception("Invalid return URL");
+            // Return URL could be malicious
+            throw new InvalidReturnUrlException();
```
sam-roberts-bjss: Still not keen on throwing the exception here. It seems like we are using it just to provide a bit of logging. If that is the case I would rather use the logger than throw an exception. Would it not be better to redirect them to the error page instead?

We could get a second opinion on this. Obviously you have done the research. Most of the small samples I've seen do not raise this exception here.

src/NHSD.BuyingCatalogue.Identity.Api/Controllers/AccountController.cs

src/NHSD.BuyingCatalogue.Identity.Api/Controllers/AccountController.cs
sam-roberts-bjss: If we don't need it, get rid of it.

src/NHSD.BuyingCatalogue.Identity.Api/Controllers/AccountController.cs
sam-roberts-bjss: We might not want to show it but we would definitely want to log it. Sometimes you might want to show an error message if it helps the user. Not sure we have a case for it yet.

src/NHSD.BuyingCatalogue.Identity.Api/Views/Shared/Error.cshtml
sam-roberts-bjss: Yeah agree. Can be a separate task if you like?

src/NHSD.BuyingCatalogue.Identity.Api/Controllers/AccountController.cs
bjss-chris: Primarily thinking from a security point-of-view – I've not actually seen a real example message yet but if it reveals details of the security implementation then we obviously wouldn't want to display it.

src/NHSD.BuyingCatalogue.Identity.Api/Views/Shared/Error.cshtml
bjss-chris: Yeah, already created a separate task.

src/NHSD.BuyingCatalogue.Identity.Api/Controllers/AccountController.cs
bjss-chris: I think this a valid use of an exception – it's an exceptional case that we can't properly handle – and I think modelling that is its real purpose, with the added benefit that we log exceptions. 

We could just redirect to the error view but I don't think that would be as clear as throwing an exception. It would also require introducing more code into an already busy method. However, happy to get another opinion or two.

src/NHSD.BuyingCatalogue.Identity.Api/Controllers/AccountController.cs
bjss-chris: As we don't have a case for it right now, I'll remove it. We can reinstate it later if required.

src/NHSD.BuyingCatalogue.Identity.Api/Controllers/AccountController.cs
bjss-chris: After thinking about this some more, I'm going to leave the `ReturnUrl == null` checks in as a malicious actor could replay the anti-forgery tokens.

src/NHSD.BuyingCatalogue.Identity.Api/Controllers/AccountController.cs
bjss-chris: As above, I'm going to leave this as is as a malicious actor could replay the anti-forgery tokens.

src/NHSD.BuyingCatalogue.Identity.Api/Controllers/AccountController.cs
bjss-chris: I removed the `var` due to our decision to prefer explicit types where it's not obvious but I'll reinstate `var` here.

PR #29, opened 2020-03-03
5146 get all organisations

RuthBerry96:


src/NHSD.BuyingCatalogue.Identity.Api/Controllers/OrganisationsController.cs
```diff
@@ -0,0 +1,45 @@
...
+    {
+        private readonly IOrganisationRepository _organisationRepository;
+
+        public OrganisationsController(IOrganisationRepository organisationRepository)
+        {
+            _organisationRepository = organisationRepository;
+        }
+
+        [HttpGet]
+        [ProducesResponseType(StatusCodes.Status400BadRequest)]
```
stuart-bass-bjss: HOw would you get anything other than a 200 response from this?

src/NHSD.BuyingCatalogue.Identity.Api/Controllers/OrganisationsController.cs
```diff
@@ -0,0 +1,45 @@
...
+        [ProducesResponseType(StatusCodes.Status204NoContent)]
+        [ProducesResponseType(StatusCodes.Status404NotFound)]
+        public async Task<ActionResult> GetAll()
+        {
+            IEnumerable<Organisation> organisationsList = await _organisationRepository.ListOrganisationsAsync();
+
+            return Ok(new GetAllOrganisationsViewModel
```
stuart-bass-bjss: Why have this view model?
IMHO, this should be Task<IEnumerable<OrganisationViewModel>>, as this is the simple case as described in the MS docs - https://docs.microsoft.com/en-us/aspnet/core/web-api/action-return-types?view=aspnetcore-3.1

tests/NHSD.BuyingCatalogue.Identity.Api.IntegrationTests/ClearDatabase.cs
```diff
@@ -0,0 +1,23 @@
...
+
+namespace NHSD.BuyingCatalogue.Identity.Api.IntegrationTests
+{
+    [Binding]
+    public sealed class ClearDatabase
+    {
+        [AfterScenario]
+        public void Clear(IConfigurationRoot config)
+        {
+            var optionsBuilder = new DbContextOptionsBuilder<ApplicationDbContext>();
```
stuart-bass-bjss: These should move into another method in the database class, which should be called from here, so the non-builder DB code is in one place.

tests/NHSD.BuyingCatalogue.Identity.Api.IntegrationTests/RegisterDependencies.cs
```diff
@@ -0,0 +1,23 @@
...
+    {
+        [BeforeScenario]
+        public void Register(IObjectContainer objectContainer)
+        {
+            var config = new ConfigurationBuilder()
+                .AddJsonFile("appsettings.json")
+                .AddEnvironmentVariables()
+                .Build();
+
+            Database.ResetUser(config.GetConnectionString("CatalogueUsersAdmin"));
```
stuart-bass-bjss: suggest this is consistent with clearing the data. Either reset the user & clear the db both before or after the run.
Thinking about it, this will get run before each test... It may be better to try registering the instance [BeforeTestRun], so it's done once per run, as opposed to per test.

src/NHSD.BuyingCatalogue.Identity.Api/Controllers/OrganisationsController.cs
RuthBerry96: GetAllOrganisationsViewModel is needed to fufil the JSON contract

RuthBerry96: GetAllOrganisationsViewModel is needed to fulfill the JSON contract below

https://buyingcatalog.visualstudio.com/Buying%20Catalogue/_sprints/taskboard/Buying%20Catalogue%20Team/Buying%20Catalogue/Iteration%2019?workitem=5201

stuart-bass-bjss: Hmm. Normally you'd just return a list of objects, not add the extra layer. I guess it's in the spec though, so fine.

PR #33, opened 2020-03-04
5167: Changed local config to work with local k8s by default

assoomro:


src/NHSD.BuyingCatalogue.Identity.Api/appsettings.json
```diff
@@ -1,5 +1,5 @@
 {
-  "issuerUrl": "null",
+  "issuerUrl": "http://docker.for.win.localhost:5102",
```
sam-roberts-bjss: How come this base address is different to all the others? e.g. host.docker.internal:5102

assoomro: There is an issue with networking when the client - in the case I tested: PB is in k8s cluster.
The client can make req. to authenticate fine, is redirected to login page as expected, however when the IS authenticates the client and sends back the token to PB, PB is unable to obtain the token. 

This 'magical' url allows for domain name resolving in and outside of containers. What I mean is
http://localhost:foo == http://docker.for.win.localhost:foo && 
http://host.docker.internal:foo == http://docker.for.win.localhost:foo
which is currently the only way to have them communicate properly.

I'm aware that this might be an issue for mac users what want to develop ISAPI locally, however there is a solution- user secrets where they would override the issuerUrl from appsettings.json to be:
http://docker.for.mac.localhost:5102

PR #34, opened 2020-03-04
3540 login validation

bjss-chris:
Adds validation to the login view.

AB#5284


src/NHSD.BuyingCatalogue.Identity.Api/ViewModels/Account/LoginViewModel.cs
```diff
@@ -5,12 +5,12 @@ namespace NHSD.BuyingCatalogue.Identity.Api.ViewModels.Account
 {
     public sealed class LoginViewModel
     {
-        [Required]
+        [Required(ErrorMessage = "Enter your password")]
         public string Password { get; set; }
```
sam-roberts-bjss: We could add a data type here to represent a password. That way we don't need to specify the type within the actual html. Also I have seen this done on a few samples. Just one thing I missed when adding the initial login view.

src/NHSD.BuyingCatalogue.Identity.Api/Views/Account/Login.cshtml
```diff
@@ -15,20 +15,20 @@
...
 
                         <div data-test-id="email-field">
                             <div data-test-id="text-field-input" class="">
                                 <div class="nhsuk-form-group">
                                     <label asp-for="Username" class="nhsuk-label">
                                         Email address
                                     </label>
-                                    <input asp-for="Username" class="nhsuk-input"/>
+                                    <input asp-for="Username" class="nhsuk-input" />
+                                    <span asp-validation-for="Username" class="text-danger"></span>
```
sam-roberts-bjss: As discussed the class used here is a bootstrap class which we currently don't reference / use within the application. I think it would be clearer here just to remove them. We should be getting a user story to decide how the login page should look with errors.

PR #35, opened 2020-03-05
5353: Making 'requirePKCE' configurable

assoomro:


src/NHSD.BuyingCatalogue.Identity.Api/Settings/ClientSettingCollection.cs
```diff
@@ -16,6 +16,7 @@ public sealed class ClientSetting
         public string AllowedGrantTypes { get; set; }
         public bool AllowOfflineAccess { get; set; }
         public bool RequireClientSecret { get; set; }
+        public bool RequirePKCE { get; set; }
```
bjss-chris: I think `RequirePkce` would probably be more in keeping with the .NET casing/naming guidelines.

PR #36, opened 2020-03-06
3540 login unit tests

bjss-chris:
AB#5343

### DIFF
--- a/tests/NHSD.BuyingCatalogue.Identity.Api.UnitTests/Controllers/AccountControllerTests.cs
+++ b/tests/NHSD.BuyingCatalogue.Identity.Api.UnitTests/Controllers/AccountControllerTests.cs
@@ -1,18 +1,289 @@
 ﻿using System;
+using System.Linq;
 using System.Threading.Tasks;
 using FluentAssertions;
+using IdentityServer4.Events;
 using IdentityServer4.Models;
 using Microsoft.AspNetCore.Mvc;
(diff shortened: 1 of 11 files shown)


tests/NHSD.BuyingCatalogue.Identity.Api.UnitTests/Controllers/AccountControllerTests.cs
```diff
@@ -1,18 +1,198 @@
...
+        [Test]
+        public async Task Login_LoginViewModel_FailedSignIn_ReturnsExpectedView()
+        {
+            using var controller = new AccountControllerBuilder()
+                .WithSignInResult(new SignInResult(false))
+                .Build();
+
+            var result = await controller.Login(new LoginViewModel()) as ViewResult;
+
+            Assert.NotNull(result);
```
sam-roberts-bjss: Any reason for the different assertion here? For consistency reasons I would try stick with the fluent assertions syntax which would be something like `result.Should().NotBeNull();`. This library tends to provide better test failure messages.

bjss-chris: R# currently isn't clever enough to identify the fluent version as a not null assertion. Therefore, I decided to use the `Assert.NotNull` version to avoid generating and suppressing additional R# warnings.

PR #42, opened 2020-03-10
4999 create pipeline

roses1089:
Merging changes to ensure the image used within the ARC (Azure Container Registry) is the latest version


azure-pipelines.yml
```diff
@@ -0,0 +1,96 @@
...
+      arguments: '-v n'
+    enabled: false
+
+  - script: |
+      docker-compose down -v
+    displayName: 'Docker down'
+
+  - task: DockerInstaller@0
+    inputs:
+      dockerVersion: '18.09.9'
```
assoomro: I think you must've forgotten to replace the version here with the variable you declared on top. 
instead of `dockerVersion: '18.09.9'`
I think it should be: `dockerVersion: $(dockerVersion)`

azure-pipelines.yml
```diff
@@ -0,0 +1,96 @@
...
+
+  pool:
+    vmImage: 'ubuntu-latest'
+
+  steps:
+
+  - task: UseDotNet@2
+    displayName: 'Use DotNet Core 3.1'
+    inputs:
+      version: 3.1.x
```
assoomro: I think you must've forgotten to replace the version here with the variable you declared on top.
instead of `version: 3.1.x`
I think it should be: `version: $(dotnetVersion)`

PR #43, opened 2020-03-11
5282: Add organisation claims

sam-roberts-bjss:


samples/NHSD.BuyingCatalogue.Identity.Api.SampleClient/Program.cs
```diff
@@ -19,7 +19,7 @@ private static async Task Main()
                 Console.WriteLine(discoveryDocument.Error);
```
CallumBradburyBJSS: Are there no changes in this file?

samples/NHSD.BuyingCatalogue.Identity.Api.SampleMvcClient/Views/Home/Privacy.cshtml
```diff
@@ -5,3 +5,18 @@
 
 <p>You are successfully logged in</p>
 <p id="sampleResourceResult">@ViewBag.Response</p>
+
```
CallumBradburyBJSS: If we have a separate page for this, is there a reason to add it here too?

src/NHSD.BuyingCatalogue.Identity.Api/Controllers/AccountController.cs
```diff
@@ -63,6 +62,8 @@ public async Task<IActionResult> Login(LoginViewModel viewModel)
             var returnUrl = viewModel.ReturnUrl.ToString();
 
             if (signInResult.IsTrustedReturnUrl)
```
CallumBradburyBJSS: Not related to this PR, but I thought we agreed to always have braces around if statements

src/NHSD.BuyingCatalogue.Identity.Api/Repositories/UserRepository.cs
```diff
@@ -0,0 +1,19 @@
...
+    public sealed class UserRepository : IUserRepository
+    {
+        private readonly UserManager<ApplicationUser> _userManager;
+
+        public UserRepository(UserManager<ApplicationUser> userManager)
+        {
+            _userManager = userManager ?? throw new ArgumentNullException(nameof(userManager));
+        }
+
+        public async Task<ApplicationUser> FindByIdAsync(string userId) => await _userManager.FindByIdAsync(userId);
```
CallumBradburyBJSS: This isn't really something I'd change, but it's a bit weird to me how we await userManager here, and then the caller has to await us awaiting the userManager. Functionally it seems like we could just return _userManager.FindByIdAsync without awaiting it, and that would save on one whole await call, letting the caller await the _userManager itself.

I think that would probably go against some sort of coding practice, but it does feel a bit odd to me when we're essentially nesting awaits for no reason.

src/NHSD.BuyingCatalogue.Identity.Api/Services/ProfileService.cs
```diff
@@ -0,0 +1,130 @@
...
+using NHSD.BuyingCatalogue.Identity.Api.Models;
+using NHSD.BuyingCatalogue.Identity.Api.Repositories;
+
+namespace NHSD.BuyingCatalogue.Identity.Api.Services
+{
+    public sealed class ProfileService : IProfileService
+    {
+        private readonly IUserRepository _applicationUserRepository;
+
+        private static readonly IDictionary<string, IEnumerable<Claim>> _organisationFunctionClaims =
```
CallumBradburyBJSS: Should this be configurable instead of hard coded?

src/NHSD.BuyingCatalogue.Identity.Api/Repositories/UserRepository.cs
```diff
@@ -0,0 +1,19 @@
...
+
+namespace NHSD.BuyingCatalogue.Identity.Api.Repositories
+{
+    public sealed class UserRepository : IUserRepository
+    {
+        private readonly UserManager<ApplicationUser> _userManager;
+
+        public UserRepository(UserManager<ApplicationUser> userManager)
+        {
+            _userManager = userManager ?? throw new ArgumentNullException(nameof(userManager));
```
CallumBradburyBJSS: .ThrowIfNull()?

src/NHSD.BuyingCatalogue.Identity.Api/Services/ProfileService.cs
```diff
@@ -0,0 +1,130 @@
...
+            new Dictionary<string, IEnumerable<Claim>>
+            {
+                { "Authority", new List<Claim> { new Claim(ApplicationClaimTypes.Organisation, "Manage") } }
+            };
+
+        public ProfileService(IUserRepository applicationUserRepository)
+        {
+            _applicationUserRepository = applicationUserRepository ?? throw new ArgumentNullException(nameof(applicationUserRepository));
```
CallumBradburyBJSS: .ThrowIfNull()?

tests/NHSD.BuyingCatalogue.Identity.Api.UnitTests/Services/ProfileServiceTests.cs
```diff
@@ -0,0 +1,377 @@
...
+                .Build();
+
+            var profileDataRequestContext = ProfileDataRequestContextBuilder
+                .Create()
+                .WithSubjectId(expectedApplicationUser.Id)
+                .Build();
+
+            await sut.GetProfileDataAsync(profileDataRequestContext);
+
+            var expected = new List<KeyValuePair<string, string>>
```
CallumBradburyBJSS: using KeyValuePair here is a bit nasty, it adds a lot of noise. We could do something like:

var expected = new List<(string, string)>{
  (Subject, expectedApplicationUser.Id),
  (PreferredUserName, expectedApplicationUser.UserName)
};

var actual = blah.Select(item => (item.Type, item.Value))
actual.Should().BeEquivilentTo(expected);

tests/NHSD.BuyingCatalogue.Identity.Api.UnitTests/Services/ProfileServiceTests.cs
```diff
@@ -0,0 +1,377 @@
...
+                .Create()
+                .WithId(expectedUserId)
+                .WithUserName(expectedUserName)
+                .Build();
+
+            Mock<IUserRepository> applicationUserRepositoryMock = new Mock<IUserRepository>();
+            applicationUserRepositoryMock.Setup(x => x.FindByIdAsync(It.IsAny<string>()))
+                .ReturnsAsync(expectedApplicationUser);
+
+            var sut = ProfileServiceBuilder
```
CallumBradburyBJSS: I hate the variable name 'sut', profileService would be better. We know that's the system under test anyway, and it makes it more readable I think

src/NHSD.BuyingCatalogue.Identity.Api/Repositories/UserRepository.cs
bjss-chris: This is better!

src/NHSD.BuyingCatalogue.Identity.Api/Controllers/AccountController.cs
bjss-chris: We agreed not to strictly enforce one way or the other, except when the statement block spans multiple lines in which case it should always have braces.

Perhaps I should start putting a document together rather than waiting for our next meeting...

samples/NHSD.BuyingCatalogue.Identity.Api.SampleClient/Program.cs
sam-roberts-bjss: This could have been a merge conflict. Looks like tabbing.

samples/NHSD.BuyingCatalogue.Identity.Api.SampleMvcClient/Views/Home/Privacy.cshtml
sam-roberts-bjss: There are two sets of claims. ID claims & user claims. Hence the introduction of a new page as need to actually go to the Identity server to retrieve the user claims .

src/NHSD.BuyingCatalogue.Identity.Api/Controllers/AccountController.cs

src/NHSD.BuyingCatalogue.Identity.Api/Repositories/UserRepository.cs
sam-roberts-bjss: Good point & you are correct. Most guidance I have seen suggests awaiting here rather than just returning the task. Quick look up found the following resource:

[Async await advice](https://blog.stephencleary.com/2016/12/eliding-async-await.html)

src/NHSD.BuyingCatalogue.Identity.Api/Services/ProfileService.cs
sam-roberts-bjss: I think it should be in the database & it very much seems like a role attribute. Something I need to follow up with @michaelpace-bjss. These claims will relate to our authentication policies so I would want to be very careful if putting them in configuration. Usually it would require a code change to introduce a new policy along with a new claim.

tests/NHSD.BuyingCatalogue.Identity.Api.UnitTests/Services/ProfileServiceTests.cs
sam-roberts-bjss: I thought this would help with readability for other developers. You know to always look for sut across all your tests & don't have to deal with the different names across multiple test fixtures.

[System under test, actual & expected](https://medium.com/@lorenzen.jacob/unit-test-naming-conventions-875786f53790)

tests/NHSD.BuyingCatalogue.Identity.Api.UnitTests/Services/ProfileServiceTests.cs
sam-roberts-bjss: Yeah good shout on this one.

PR #45, opened 2020-03-11
5146: Organisation DB restructure

RuthBerry96:


src/NHSD.BuyingCatalogue.Identity.UserDatabase/Tables/Organisations.sql
```diff
@@ -1,7 +1,8 @@
 ﻿CREATE TABLE [dbo].[Organisations]
 (
-	[Id] UNIQUEIDENTIFIER NOT NULL PRIMARY KEY, 
-    [Name] VARCHAR(255) NOT NULL, 
-    [OdsCode] VARCHAR(8) NULL, 
-    [LastUpdated] DATETIME2 NOT NULL
-)
+	Id UNIQUEIDENTIFIER NOT NULL,
```
sam-roberts-bjss: The name column has brackets but the others don't. Could we make this consistent?

src/NHSD.BuyingCatalogue.Identity.UserDatabase/Tables/Organisations.sql
```diff
@@ -1,7 +1,8 @@
...
+    [Name] NVARCHAR(255) NOT NULL,
+    OdsCode NVARCHAR(8) NULL,
+    LastUpdated DATETIME2 NOT NULL DEFAULT GETUTCDATE(),
```
sam-roberts-bjss: I believe the last updated column should be a DATETIME2 (7)?

sam-roberts-bjss: Not sure about the default GETUTCDATE() as I think by leaving this out ensures the application code has to specify a datetime & would give more flexibility.

src/NHSD.BuyingCatalogue.Identity.UserDatabase/Tables/Organisations.sql
```diff
@@ -1,7 +1,8 @@
...
+    CONSTRAINT PK_Organisations PRIMARY KEY NONCLUSTERED (Id)
```
sam-roberts-bjss: Should this constraint have ASC on the end? Just compared a PK constraint with the Solutions database script.

src/NHSD.BuyingCatalogue.Identity.UserDatabase/Tables/Organisations.sql
bjss-chris: The reason for the brackets around `name` is just because VS will highlight it as a reserved word (even though it isn't). We can remove the brackets here but I suggest we just leave them in this case as there could be other cases if we ever use a reserved word for a column name but hopefully we will avoid that... However, in all other cases DDL looks better without square brackets in my opinion.

src/NHSD.BuyingCatalogue.Identity.UserDatabase/Tables/Organisations.sql
bjss-chris: The default precision is 7 so `datetime2` and `datetime(7)` are equivalent. However, it certainly wouldn't be a problem to use the latter for consistency – @RuthBerry96 can make the call on that one.

src/NHSD.BuyingCatalogue.Identity.UserDatabase/Tables/Organisations.sql
bjss-chris: `ASC` is the default here so omitting it just avoids some unnecessary clutter and I believe the key and index definitions follow this convention elsewhere within the same DB project.

src/NHSD.BuyingCatalogue.Identity.UserDatabase/Tables/Organisations.sql
bjss-chris: As `LastUpdated` will always be the current date on creation this suggestion was just to avoid the need to provide the current date on creation. Removing the default wouldn't stop `LastUpdated` from being omitted in a subsequent update but it's not going to cause any problems either so can be removed if preferred.

sam-roberts-bjss:
Yeah good shout on the clustered index for the name. I'm not sure we need all those properties on the index as most of them will be the default. @RuthBerry96 could you check this please? 

The idea to not introduce all the fields at once is to make the tasks more manageable. As it stands the Front end only needed the pieces of data that have been added. Once we get to the data task to import the Organisation the plan was to add the additional properties then.

--------------------

After a quick chat is sounds like you will be moving onto the data task pretty quickly. So could you include the address column please @RuthBerry96.

src/NHSD.BuyingCatalogue.Identity.UserDatabase/Tables/Organisations.sql
RuthBerry96: Changed to DateTime2(7) to remain constant with the buyingcatalogue

src/NHSD.BuyingCatalogue.Identity.UserDatabase/Tables/Organisations.sql
```diff
@@ -1,8 +1,10 @@
 ﻿CREATE TABLE [dbo].[Organisations]
 (
 	Id UNIQUEIDENTIFIER NOT NULL,
-    [Name] NVARCHAR(255) NOT NULL,
+    [Name] NVARCHAR(255) NOT NULL,    
+    [Address] NVARCHAR(255) NULL,
```
sam-roberts-bjss: The address should probably be max. There are questions of whether or not this should be in its own table though. One for @michaelpace-bjss I think?

PR #46, opened 2020-03-11
3540: Fixing Error Callback

CallumBradburyBJSS:


src/NHSD.BuyingCatalogue.Identity.Api/Controllers/ErrorController.cs
```diff
@@ -0,0 +1,31 @@
+﻿using System.Threading.Tasks;
+using IdentityServer4.Services;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.Extensions.Logging;
+
+namespace NHSD.BuyingCatalogue.Identity.Api.Controllers
+{
+    public class ErrorController : Controller
```
sam-roberts-bjss: Sealed?

src/NHSD.BuyingCatalogue.Identity.Api/Controllers/ErrorController.cs
```diff
@@ -0,0 +1,31 @@
...
+            ILogger<ErrorController> logger)
+        {
+            _interactionService = interactionService;
+            _logger = logger;
+        }
+
+        public async Task<IActionResult> Index(string errorId)
+        {
+            var message = await _interactionService.GetErrorContextAsync(errorId);
+            _logger.Log(LogLevel.Error, "Request for client {0} failed with error code {1}: {2}",
```
sam-roberts-bjss: One suggestion here is to call _logger.LogInformation rather than the specific implementation to keep it consistent with the rest of the product suite. Also review the use of passing the message into the logger as an object. This way Serilog can get involved & JSON stringify the data. I'm not quite sure of the syntax here, hence the request for review.

src/NHSD.BuyingCatalogue.Identity.Api/Controllers/ErrorController.cs
```diff
@@ -21,7 +21,7 @@ public class ErrorController : Controller
         public async Task<IActionResult> Index(string errorId)
         {
             var message = await _interactionService.GetErrorContextAsync(errorId);
-            _logger.Log(LogLevel.Error, "Request for client {0} failed with error code {1}: {2}",
+            _logger.Log(LogLevel.Error, "Request for client {ClientId} failed with error code {Error}: {ErrorDescription}",
```
sam-roberts-bjss: Did we agree here to use the _logger.LogError extension method rather than the concrete implementation?

tests/NHSD.BuyingCatalogue.Identity.Api.UnitTests/Controllers/ErrorControllerTests.cs
```diff
@@ -19,46 +16,32 @@ public sealed class ErrorControllerTests
         [Test]
         public async Task Index_WithId_DisplaysErrorView()
         {
-            var context = new TestContext();
-            var res = await context.Controller.Index("testId") as ViewResult;
+            using var controller = new ErrorControllerBuilder().Build();
+
+            var res = await controller.Index("testId") as ViewResult;
```
sam-roberts-bjss: What does res mean? Could this be made clearer?

CallumBradburyBJSS:
> I like the use of the builder but it looks like it combines two different paradigms with the builder pattern and the test context. The builder class has pretty much become the test context. The builder class has one responsibility to 'build' an object with certain defaults. It was only used as a mitigating impact when changes are made to the construction of an object in the future.
> The set of tests here are that small do we even need a test context or a 'builder' here?

I think it's worth having something to abstract out the construction, whether that's a builder or a test context, but I'm pretty happy combining the two paradigms if that gives the cleanest results with regards to the readability of the tests.

sam-roberts-bjss:
Unfortunately I don't think merging the two approaches is the right way to go. The "builder" pattern here is meant to be a simple object creation with sensible defaults & reduce the impact of dependency changes to the unit tests in the future.

We both want a good solution & I am all for easy to read unit tests, but I think at the moment consistency is key. Keep is simple and consistent to start with & can expand on it later. Lets just get something delivered in this sprint.

PR #55, opened 2020-03-16
5147: Organisation additional columns

RuthBerry96:


src/NHSD.BuyingCatalogue.Organisations.Api/Controllers/OrganisationsController.cs
```diff
@@ -30,12 +30,26 @@ public async Task<ActionResult> GetAllAsync()
...
-                })
+                Organisations = organisationsList?.Select(x =>
+                    new OrganisationViewModel
+                    {
+                        OrganisationId = x.Id,
+                        Name = x.Name,
+                        OdsCode = x.OdsCode,
+                        PrimaryRoleId = x.PrimaryRoleId,
+                        CatalogueAgreementSigned = x.CatalogueAgreementSigned,
+                        Location = x.Location == null ? null : new LocationViewModel
```
sam-roberts-bjss: The name of the Location property confused me slightly. It sounds more like a GPS location rather than a postal address. The agreed contract with the Frontend does method the property 'address', however it is in a different format. I spoke to @alibombaye & agreed the new JSON contract

src/NHSD.BuyingCatalogue.Organisations.Api/Models/Organisation.cs
```diff
@@ -1,22 +1,37 @@
...
+        public Address Location
         {
-            Id = id;
-            Name = name;
-            OdsCode = odsCode;
+            get
+            {
+                if (Address != null)
+                {
+                    _location = JsonConvert.DeserializeObject<Address>(Address);
```
sam-roberts-bjss: Ideally the domain model should have no idea how the data is stored. Adding the JSON and backing field will couple the domain with our infrastructure.

Here is a resource for reference (we are using an Address object rather than a dictionary).
[Use Json conversion with EF Core](https://www.jerriepelser.com/blog/store-dictionary-as-json-using-ef-core-21/)

tests/NHSD.BuyingCatalogue.Identity.Api.IntegrationTests/Features/Organisations GetById 5147.feature
```diff
@@ -16,6 +16,16 @@ Scenario: 1. Get the details of a single organisation
 	Then a response with status code 200 is returned
 	And the string value of name is Organisation 1
 	And the string value of odsCode is Ods 1
+	And the string value of primaryRoleId is ID 1
```
sam-roberts-bjss: I compared this section with the GET all organisation tests & I wondered why the checking of the result is different. Do you think it would be a good idea (if possible) to utilize the approached used in the get all feature file?

tests/NHSD.BuyingCatalogue.Identity.Api.IntegrationTests/Steps/Common/CommonSteps.cs
```diff
@@ -21,10 +22,19 @@ public void AResponseIsReturned(int code)
         }
 
         [Then(@"the string value of (.*) is (.*)")]
-        public async Task ThenTheStringValueOfNameIs(string section, string value)
+        public async Task ThenTheStringValueIs(string section, string value)
         {
             var content = await _response.ReadBody();
-            content.Value<string>(section).Should().Be(value);
+
+            var token = content.SelectToken(section) as JValue;
```
sam-roberts-bjss: In the Buying Catalogue API we added some additional checking here to make sure the section is contained within the response. It improved the test failure error message as before we would just get an exception.

Here is an example:

`
content.Should().NotBeNull();
content.ToString().Should().Contain($"\"{section}\"");
`

tests/NHSD.BuyingCatalogue.Identity.Api.UnitTests/OrganisationControllerTests.cs
```diff
@@ -13,8 +14,32 @@
 
 namespace NHSD.BuyingCatalogue.Identity.Api.UnitTests
 {
-    public sealed class OrganisationControllerTests 
+    public sealed class OrganisationControllerTests
     {
+        private Address address1 = new Address
```
sam-roberts-bjss: The address fields here should be marked as readonly as this will ensure the address data cannot be pointed to a new value. Also the convention for member variables is to prefix them with an underscore.

tests/NHSD.BuyingCatalogue.Identity.Api.IntegrationTests/Steps/OrganisationsSteps.cs
```diff
@@ -73,12 +85,48 @@ public async Task ThenTheOrganisationsListIsReturnedWithTheFollowingValues(Table
 
             var organisations = (await _response.ReadBody()).SelectToken("organisations").Select(t => new
             {
-                Name = t.SelectToken("name").ToString(), OdsCode = t.SelectToken("odsCode").ToString()
+                Name = t.SelectToken("name").ToString(),
```
sam-roberts-bjss: Could this section be refactored? It looks the same as the change below. This would reduce the amount of code & reduce the duplication of token constants, so if one ever changed we don't need to know to change it else where.

tests/NHSD.BuyingCatalogue.Identity.Api.IntegrationTests/Steps/OrganisationsSteps.cs
```diff
@@ -73,12 +85,48 @@ public async Task ThenTheOrganisationsListIsReturnedWithTheFollowingValues(Table
...
 
             organisations.Should().BeEquivalentTo(expectedOrganisations, options => options.WithStrictOrdering());
         }
 
+        [Then(@"the Organisation is returned with the following values")]
+        public async Task ThenTheOrganisationIsReturnedWithTheFollowingValues(Table table)
+        {
+            var expectedOrganisation = table.CreateSet<OrganisationTable>().FirstOrDefault();
+
+            var org = await _response.ReadBody();
```
sam-roberts-bjss: Try to make variable names appropriate to the object you are initialising to e.g. responseBody. So when reading the code one doesn't have to context switch & will hopefully follow it better.

tests/NHSD.BuyingCatalogue.Identity.Api.UnitTests/OrganisationControllerTests.cs
```diff
@@ -64,18 +101,54 @@ public async Task GetAllAsync_SingleOrganisationExists_ReturnsTheOrganisation(st
...
+            }
         }
 
         [Test]
         public async Task GetAllAsync_ListOfOrganisationsExist_ReturnsTheOrganisations()
         {
-            var org1 = new Organisation(new Guid(), "Organisation 1", "ODS 1");
-            var org2 = new Organisation(new Guid(), "Organisation 2", "ODS 2");
-            var org3 = new Organisation(new Guid(), "Organisation 3", "ODS 3");
+            var org1 = new Organisation
```
sam-roberts-bjss: One suggestion here would be to reduce the amount of code by introducing the builder pattern for the organisation class. This could contain sensible defaults. If we add another property to the organisation class in the future it could potentially effect multiple tests. With the builder pattern it could potentially mitigate this impact with a sensible default.

tests/NHSD.BuyingCatalogue.Identity.Api.UnitTests/OrganisationControllerTests.cs
```diff
@@ -94,12 +167,21 @@ public async Task GetAllAsync_ListOfOrganisationsExist_ReturnsTheOrganisations()
 
             organisationList[0].Name.Should().Be(org1.Name);
```
sam-roberts-bjss: Rather than compare each property, could the code compare object to object? e.g. `organisationList[0].Should().Be(org1)`. This will make the test easier to read by reducing the amount of comparison code.

tests/NHSD.BuyingCatalogue.Identity.Api.UnitTests/OrganisationControllerTests.cs
```diff
@@ -146,12 +236,35 @@ public async Task GetIdByAsync_OrganisationExists_ReturnsTheOrganisation()
             result.Should().BeOfType<OkObjectResult>();
             var objectResult = result as OkObjectResult;
 
-            objectResult.Value.Should().BeEquivalentTo(new OrganisationViewModel
+            objectResult.Value.Should().BeEquivalentTo(organisation,
+                conf => conf.Excluding(c => c.Id).Excluding(c => c.Address).Excluding(c => c.LastUpdated));
```
sam-roberts-bjss: Can you review the use of excluding the address from the comparison? An address is assigned to the organisation object, should this not be included in the assert?

tests/NHSD.BuyingCatalogue.Identity.Api.UnitTests/OrganisationControllerTests.cs
```diff
@@ -146,12 +236,35 @@ public async Task GetIdByAsync_OrganisationExists_ReturnsTheOrganisation()
...
+                .Create()
+                .WithGetOrganisation(organisation)
+                .Build();
+
+            var result = await controller.GetByIdAsync(organisation.Id);
+
+            result.Should().BeOfType<OkObjectResult>();
+            var objectResult = result as OkObjectResult;
+
+            objectResult.Value.Should().BeEquivalentTo(organisation,
```
sam-roberts-bjss: Again here please check the excluding of the address property. The assert should ensure the address is null seen as we are setting the `Address `property to `null`.

tests/NHSD.BuyingCatalogue.Identity.Api.IntegrationTests/Steps/OrganisationsSteps.cs
```diff
@@ -150,6 +125,25 @@ public async Task GivenAnOrganisationWithNameOrganisationDoesNotExist(string org
             organisations.Should().BeNull();
         }
 
+        private object CreateOrganisation(JToken token)
```
sam-roberts-bjss: I believe this method could be static? It will give a very slight performance benefit & hopefully give a bit of context to other developers within the code base, indicating this is not an instance method.

tests/NHSD.BuyingCatalogue.Identity.Api.UnitTests/OrganisationControllerTests.cs
```diff
@@ -117,34 +117,11 @@ public async Task GetAllAsync_SingleOrganisationExists_ReturnsTheOrganisation(st
...
-                Id = Guid.NewGuid(),
-                Name = "Organisation 3",
-                OdsCode = "ODS 3",
-                PrimaryRoleId = "ID 3",
-                CatalogueAgreementSigned = true,
-                Address = null
-            };
+            var org1 = OrganisationBuilder.Create(1).WithAddress(_address1).Build();
+
+            var org2 = OrganisationBuilder.Create(2).WithCatalogueAgreementSigned(true).WithAddress(_address2).Build();
```
sam-roberts-bjss: There looks to be more calls to `WithCatalogueAgreementSigned(true)`. Would it simplify the code to switch the default to `true` & just create one organisation with a false Catalogue Agreement Signed?

tests/NHSD.BuyingCatalogue.Identity.Api.UnitTests/OrganisationControllerTests.cs
```diff
@@ -36,17 +60,30 @@ public async Task GetAllAsync_NoOrganisationsExist_EmptyResultIsReturned()
...
+        public async Task GetAllAsync_SingleOrganisationExists_ReturnsTheOrganisation(string name, string ods, string primaryRoleId, bool catalogueAgreementSigned, bool hasAddress)
         {
             var organisationId = Guid.NewGuid();
 
             using var controller = OrganisationControllerBuilder
                 .Create()
-                .WithListOrganisation(new List<Organisation>(){new Organisation(organisationId, name, ods)})
+                .WithListOrganisation(new List<Organisation>()
+                {
+                    new Organisation()
```
sam-roberts-bjss: Just to reiterate what I said on the phone. The OrganisationBuilder class could be used here, so that the code contains one area for creating a new Organisation objects. If the Organisation class changes, hopefully there will only be the builder to alter, rather than multiple tests.

PR #58, opened 2020-03-17
5147: Added canned data for users

CallumBradburyBJSS:


src/NHSD.BuyingCatalogue.Organisations.Api/ViewModels/OrganisationUsers/GetAllOrganisationUsersViewModel.cs
```diff
@@ -0,0 +1,10 @@
+﻿using Newtonsoft.Json;
+
+namespace NHSD.BuyingCatalogue.Organisations.Api.ViewModels.OrganisationUsers
+{
+    public sealed class GetAllOrganisationUsersViewModel
+    {
+        [JsonProperty("users")]
```
bjss-chris: Do we need `JsonProperty` the attribute here?

src/NHSD.BuyingCatalogue.Organisations.Api/ViewModels/OrganisationUsers/OrganisationUserViewModel.cs
```diff
@@ -0,0 +1,22 @@
...
+        [JsonProperty("name")]
+        public string Name { get; set; }
+        
+        [JsonProperty("phoneNumber")]
+        public string PhoneNumber { get; set; }
+        
+        [JsonProperty("emailAddress")]
+        public string EmailAddress { get; set; }
+
+        [JsonProperty("isDisabled")]
```
bjss-chris: Do we need `JsonProperty` the attributes here?

src/NHSD.BuyingCatalogue.Organisations.Api/ViewModels/OrganisationUsers/GetAllOrganisationUsersViewModel.cs
CallumBradburyBJSS: No, I just like having it incase someone does a rename sometime in the future, and misses the json property. I think as we know for sure the contract we're implementing, we should probably implicitly state that is what we want.

bjss-chris: Yeah, it makes sense for renames – but it should be `JsonPropertyName` as we're currently using the `System.Text` serializer although I don't know if that's intentional or accidental (it's the default JSON serializer now).

CallumBradburyBJSS: I have removed the JsonProperty attributes for now, to get the Canned Data in, but will reintroduce them later

PR #64, opened 2020-03-18
Add organization data

bjss-chris:
AB#5200


src/NHSD.BuyingCatalogue.Identity.UserDatabase/Tables/Organisations.sql
```diff
@@ -1,12 +1,12 @@
...
-    PrimaryRoleId NVARCHAR(8) NULL,
-    CatalogueAgreementSigned BIT NOT NULL,
-    LastUpdated DATETIME2(7) NOT NULL,
+	OrganisationId uniqueidentifier NOT NULL,
+    [Name] nvarchar(255) NOT NULL,
+    [Address] nvarchar(max) NULL,
+    OdsCode nvarchar(8) NULL,
+    PrimaryRoleId nvarchar(8) NULL,
+    CatalogueAgreementSigned bit DEFAULT 0 NOT NULL,
+    LastUpdated datetime2(7) DEFAULT GETUTCDATE() NOT NULL,
```
sam-roberts-bjss: Ruth & I agreed to not include the 'DEFAULT GETUTCDATE()'. I assume this would effect all your scripts if this was removed? Or the scripts would have to be changed to add a last updated date & time?

src/NHSD.BuyingCatalogue.Identity.UserDatabase/Deployment/CreateTestUsers.sql
```diff
@@ -1,11 +1,18 @@
 ﻿DECLARE @aliceEmail AS nvarchar(50) = N'AliceSmith@email.com';
 DECLARE @bobEmail AS nvarchar(50) = N'BobSmith@email.com';
 
-IF NOT EXISTS(
+IF '$(ASPNETCORE_ENVIRONMENT)' <> 'Production'
```
sam-roberts-bjss: No problems with this change. Just had a thought whether or not it would be useful to do something like the appsettings.development.json file. So we only load the SQL file depending on the `ASPNETCORE_ENVIRONMENT` environment variable. This would remove the if condition & keep the concerns of development and production separate. What do you think? Definitely not saying do this now.

docker-compose.yml
```diff
@@ -10,6 +10,7 @@ services:
     networks:
       - nhsd_bcnetwork
     environment:
+      - ASPNETCORE_ENVIRONMENT=Development
```
sam-roberts-bjss: Will we need to make any changes to the helm charts related to this alteration?

src/NHSD.BuyingCatalogue.Identity.UserDatabase/Deployment/CreateTestUsers.sql
bjss-chris: That makes sense although there's no easy way to do that that I'm aware of. I think we'd need to play around with the build config and make sure we pass the environment variable to MSBuild.

I guess we can leave like this for now and create a new task for that. Although it's not elegant at least this will prevent the test users from being created in production for now.

docker-compose.yml
bjss-chris: I have no idea as I no nothing about helm. Can @assoomro or @stuart-bass-bjss help?

src/NHSD.BuyingCatalogue.Identity.UserDatabase/Tables/Organisations.sql
bjss-chris: Yes, that's correct. All the scripts would need to be changed to set `LastUpdated` to `GETUTCDATE()` and `CatalogueAgreementSigned` to `0` (or `1`), which seems somewhat unnecessary when we can set an acceptable default.

In these two cases we can easily set an acceptable default and it also means it will avoid any issues if we still have any environments where these two columns don't yet exist in the database (as a new `NOT NULL` column without a `DEFAULT` obviously isn't a good combo if any data already exists).

I'm not sure we really gain anything by omitting the defaults.

docker-compose.yml
assoomro: We'll need to add that env. var to the helm chart. I suggest we raise a task for this to keep track of it.

bjss-chris: I've added a platform [task (5521)](https://dev.azure.com/BuyingCatalog/Buying%20Catalogue/_workitems/edit/5521) to the story. Thanks @assoomro.

src/NHSD.BuyingCatalogue.Identity.UserDatabase/Tables/Organisations.sql
stuart-bass-bjss: Defaults are easier to deal with in terms of applying an upgrade via dacpac to existing tables. My vote is for keeping them.

sam-roberts-bjss: Sounds good 👍

src/NHSD.BuyingCatalogue.Identity.UserDatabase/Deployment/CreateTestUsers.sql
stuart-bass-bjss: Is it possible instead to have this as something like INSERT_TEST_DATA, so it's clear what the intention is of the sqlcmd variable?
Also, might be worth amending the user guard, so if any are set it doesn't insert them.

sam-roberts-bjss: I was thinking something could be done within the `configure-db.sh` to look for a file with the appropriate environment variable value. But sounds like Stuart would like to change the context of the variable anyways.

bjss-chris: I just stuck with `ASPNETCORE_ENVIRONMENT` as we already had that for the .NET environments but I did originally think an additional environment variable might have been preferable, particularly as the DB isn't an ASP.NET Core environment. Therefore, I'm happy to go with Stuart's suggestion of INSERT_TEST_DATA.

@stuart-bass-bjss can you clarify what you mean by amending the guard? Currently, the test users won't be inserted if they already exist.

@sam-roberts-bjss when we deploy the dacpac using DacFx we shouldn't have any need for `configure-db.sh`. Ideally I'd like to look into whether we can tweak the docker build to make use of DacFx at some point, if time permits (tech debt).

stuart-bass-bjss: I meant to amend the guard, so the test users aren't inserted if any users exist.

bjss-chris: Ah, that's fine. That guard is already there.

bjss-chris: Or perhaps not. To clarify: do mean any *test* users or *any* users?

PR #65, opened 2020-03-18
5146: Policy added to organisation

RuthBerry96:


src/NHSD.BuyingCatalogue.Identity.Api/Settings/ApiResourceSettingCollection.cs
```diff
@@ -15,7 +16,7 @@ public sealed class ResourceSetting
 
         public ApiResource ToResource()
         {
-            return new ApiResource(ResourceName, DisplayName);
+            return new ApiResource(ResourceName, DisplayName, new List<string>(){ ApplicationClaimTypes.Organisation });
```
sam-roberts-bjss: This needs to be configurable. During development we only hardcoded it to see if we could get it working. Not all API resources will want the Organisation scope. This will involve a change to the appsettings.json (& the development version).

src/NHSD.BuyingCatalogue.Organisations.Api/Controllers/OrganisationsController.cs
```diff
@@ -11,7 +11,7 @@
 
 namespace NHSD.BuyingCatalogue.Organisations.Api.Controllers
 {
-    [Authorize]
+    [Authorize(Policy = "CanAccessOrganisation")]
```
sam-roberts-bjss: Could we make a static class for Policy names? The duplication of the name could get out of sync. Want to introduce one place to make a change.

PR #67, opened 2020-03-18
5147: Implemented GetUsersByOrganisationId

CallumBradburyBJSS:


src/NHSD.BuyingCatalogue.Organisations.Api/Repositories/UsersRepository.cs
```diff
@@ -0,0 +1,25 @@
...
+    public class UsersRepository : IUsersRepository
+    {
+        private readonly ApplicationDbContext _context;
+
+        public UsersRepository(ApplicationDbContext context)
+        {
+            _context = context;
+        }
+
+        public async Task<IEnumerable<ApplicationUser>> GetUsersByOrganisationIdAsync(Guid id)
```
sam-roberts-bjss: Although the method name is helpful in this instance the parameter variable 'id' can be a bit vague. Do you think it would be a good idea to be explicit and specify organisationId instead?

tests/NHSD.BuyingCatalogue.Identity.Api.UnitTests/Controllers/UserControllerTests.cs
```diff
@@ -0,0 +1,95 @@
...
+        internal static void AssertEqualUsers(IEnumerable<OrganisationUserViewModel> actuals, IEnumerable<ApplicationUser> expecteds)
+        {
+            actuals.Count().Should().Be(expecteds.Count());
+
+            for (var i = 0; i < expecteds.Count(); i++)
+            {
+                var actual = actuals.ElementAt(i);
+                var expected = expecteds.ElementAt(i);
+
+                actual.PhoneNumber.Should().Be(expected.PhoneNumber);
```
sam-roberts-bjss: Rather than test each property would it be better to just have one the assert line e.g. `actual.Should().Be(expected)`? Make the code easier to read & also if new properties are added to the object the test would fail.

CallumBradburyBJSS: actual.Should().Be.Expected() doesn't play ball, because the database objects differ significantly from the viewmodels. I did try setting up so that would work, but it ended up with a huge amount of code for excluding properties that ApplicationUser has, and converting between parameter names. I don't particularly like this solution, but couldn't think of anything better. I'll ponder it some more

src/NHSD.BuyingCatalogue.Organisations.Api/Repositories/UsersRepository.cs
CallumBradburyBJSS: sure, I can do that

src/NHSD.BuyingCatalogue.Organisations.Api/Controllers/UsersController.cs
```diff
@@ -35,13 +47,28 @@ public sealed class UsersController : Controller
         };
 
         [HttpGet]
-        public ActionResult GetUsersById(Guid organisationId)
+        public async Task<ActionResult> GetUsersByOrganisationId(Guid id)
         {
-            Console.WriteLine($"Use the '{organisationId}' so the build doesn't complain.");
+            var organisation = await _organisationRepository.GetByIdAsync(id);
```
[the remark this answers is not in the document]
CallumBradburyBJSS: That's a fair point, I'm happy to change it to return an empty list if there's no use for the 404

src/NHSD.BuyingCatalogue.Organisations.Api/Controllers/UsersController.cs
```diff
@@ -35,13 +47,28 @@ public sealed class UsersController : Controller
...
+            if (organisation == null)
+            {
+                return NotFound();
+            }
+
+            var organisationUsers = await _usersRepository.GetUsersByOrganisationIdAsync(id);
+            var userViewModels = organisationUsers.Select(x => new OrganisationUserViewModel
```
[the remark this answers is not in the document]
CallumBradburyBJSS: No, because it's returning the result of a linq query, which will always provide a list. Granted that's knowledge that only the UsersRepository has, so from this class' perspective it could retrieve a null, but I'm relatively happy with the assumption that it won't.

src/NHSD.BuyingCatalogue.Organisations.Api/Controllers/UsersController.cs
```diff
@@ -35,13 +47,28 @@ public sealed class UsersController : Controller
         };
 
         [HttpGet]
-        public ActionResult GetUsersById(Guid organisationId)
+        public async Task<ActionResult> GetUsersByOrganisationId(Guid id)
```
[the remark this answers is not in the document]
CallumBradburyBJSS: Yeah I can change that

tests/NHSD.BuyingCatalogue.Identity.Api.UnitTests/Controllers/UserControllerTests.cs
sam-roberts-bjss: One suggestion would be to change the comparison. If the result is a list of viewmodels, create an expected list of viewmodels and compare those rather than two different types.

sam-roberts-bjss:
One thing I forgot to mention yesterday I didn't notice a Startup class change. Do you need to register the UserRepository for dependency injection?

sam-roberts-bjss:
Just noticed the controller unit tests live inside the associated Identity.Api test project. The tests created here are for the Organisation.Api project. Do you think it would be better to create an Organisations.Api.UnitTests project to keep concerns separate?

CallumBradburyBJSS:
> Just noticed the controller unit tests live inside the associated Identity.Api test project. The tests created here are for the Organisation.Api project. Do you think it would be better to create an Organisations.Api.UnitTests project to keep concerns separate?

I do think it would be better, but I'm not doing it as part of this changeset

tests/NHSD.BuyingCatalogue.Identity.Api.UnitTests/Controllers/UserControllerTests.cs
CallumBradburyBJSS: Nice suggestion, I've made this switch

CallumBradburyBJSS:
> Do we need to add a test to check the user repository dependency?

It'd be nice, but is proving to be a real pain to unit test due to the hard coupling to a DbContext, so is probably better suited as a Database test, or accepted as being covered by the integration tests as it's a one line proxy class. I'll add it in my integration tests task if we deem it necessary.

sam-roberts-bjss:
> > Do we need to add a test to check the user repository dependency?
> 
> It'd be nice, but is proving to be a real pain to unit test due to the hard coupling to a DbContext, so is probably better suited as a Database test, or accepted as being covered by the integration tests as it's a one line proxy class. I'll add it in my integration tests task if we deem it necessary.

Sorry I meant we usually check the UsersController dependencies to ensure that the method on the UserRepository is only called once. Nothing to do with the DbContext. Is that any clearer?

tests/NHSD.BuyingCatalogue.Identity.Api.UnitTests/Controllers/UsersControllerTests.cs
```diff
@@ -0,0 +1,76 @@
...
+            var result = await controller.GetUsersByOrganisationId(Guid.Empty) as OkObjectResult;
+            
+            result.Should().NotBeNull();
+            var viewModel = result.Value as GetAllOrganisationUsersViewModel;
+            viewModel.Should().NotBeNull();
+            
+            viewModel.Users.Should().BeEquivalentTo(users.Select(x => x.Expected));
+        }
+
+        internal static (ApplicationUser RepoUser, OrganisationUserViewModel ExpectedUser) CreateUser(bool disabled)
```
sam-roberts-bjss: Is this method used elsewhere? If not should it be private?

tests/NHSD.BuyingCatalogue.Identity.Api.UnitTests/Builders/UsersControllerBuilder.cs
```diff
@@ -33,15 +27,9 @@ internal UsersControllerBuilder SetUsers(IEnumerable<ApplicationUser> users)
...
-        {
-            _organisation = organisation;
-            return this;
-        }
-
-        internal UsersController Build()
+        internal (UsersController Controller, Mock<IUsersRepository> UserRepository) Build()
         {
-            return new UsersController(_usersRepositoryMock.Object, _organisationMock.Object);
+            return (new UsersController(_usersRepositoryMock.Object), _usersRepositoryMock);
```
sam-roberts-bjss: As briefly discussed this type of pattern I think should be presented to the rest of the team, so just putting down some of my thoughts for discussion.

In my opinion this looks like a bastardization of the builder pattern & a test context approach. It is now building two objects, rather than keeping it simple to just one. The introduction of more dependencies will increase the size of the tuple returning from the build method & could quickly get out of hand.

Even though having the mocked object within the builder makes the test code a bit neater it does lose some flexibility, meaning the test method cannot setup the methods specific to its logic. How could the code test when an exception is raised from the repository?

We all want neater & easy to read tests, so I am not dismissing this approach, just have some reservations, hence why I would like to open a discussion.

CallumBradburyBJSS: > In my opinion this looks like a bastardization of the builder pattern & a test context approach. It is now building two objects, rather than keeping it simple to just one. The introduction of more dependencies will increase the size of the tuple returning from the build method & could quickly get out of hand.

**If the tuple gets too big, I see that as an indication that the class itself is doing too much and can probably be broken down**

> Even though having the mocked object within the builder makes the test code a bit neater it does lose some flexibility, meaning the test method cannot setup the methods specific to its logic. How could the code test when an exception is raised from the repository?

**This is not true, you can do built.UserController.Setup().Throws() after you have called Build(), so it's a really clean way of modifying the default built setup to fit test specific needs**

PR #75, opened 2020-03-20
Update the readme with all current endpoints

bjss-chris:
Updated the readme with a brief description of all current endpoints and completed the section on running the integration tests.

AB#5490

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,10 +20,18 @@ This application uses **.NET core** to provide an identity service currently imp
 
 This repository uses **.NET Core**, **Nodejs** and **Docker**.
 
-It contains one endpoint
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -66,4 +76,4 @@ docker-compose down -v
 
 ### Running the Integration Tests
 
-TO BE COMPLETED
+Start the application as decribed in [running the application](#running-the-application) and run the integration tests using your preferred test runner.
```
CallumBradburyBJSS: It might be worth specifying where the integration tests live, as I don't think that's mentioned anywhere in the file, but as that's not part of the logout change I will let it slide for this review

PR #79, opened 2020-03-23
3540 flaky assertion

assoomro:
This commit should prevent the pipeline failing because of a flaky assertion.


CallumBradburyBJSS:
Am I right in thinking this change just moves the test to after the unit tests, to give docker time to spin up properly?

assoomro:
> Am I right in thinking this change just moves the test to after the unit tests, to give docker time to spin up properly?

Precisely

PR #85, opened 2020-03-25
3538: Added Policies for accessing and managing organisation users

CallumBradburyBJSS:


src/NHSD.BuyingCatalogue.Identity.Common/NHSD.BuyingCatalogue.Identity.Common.csproj
```diff
@@ -0,0 +1,7 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
```
RuthBerry96: Set up fxcop

tests/NHSD.BuyingCatalogue.Identity.Api.IntegrationTests/Features/Get Organisation Users.5147.feature
```diff
@@ -8,40 +8,50 @@ Background:
...
     And Users exist
-        | Id  | OrganisationName | FirstName | LastName | Email   | PhoneNumber | Disabled | Password |
-        | 123 | Organisation 1   | John      | Doe      | a@b.com | 01234567890 | false    | yolo     |
+        | Id  | OrganisationName | FirstName | LastName | Email             | PhoneNumber | Disabled | Password | OrganisationFunction |
+        | 123 | Organisation 1   | John      | Doe      | authority@doe.com | 01234567890 | false    | yolo     | Authority            |
+        | 234 | Organisation 1   | Jane      | Doe      | buyer@doe.com     | 01234567890 | false    | oloy     | Buyer                |
 
 @5147
 Scenario: 1. Get all of the users for an organisation
     Given an user is logged in
```
RuthBerry96: Given a authority user is logged in

CallumBradburyBJSS: That step doesn't exist anymore, right? I'm logging in as authority@doe.com which logs us in as an authority

src/NHSD.BuyingCatalogue.Identity.Api/Settings/IdentityResourceSetting.cs
```diff
@@ -1,5 +1,5 @@
-﻿using IdentityServer4.Models;
-using NHSD.BuyingCatalogue.Identity.Api.Constants;
+﻿using BuyingCatalogue.Identity.Common.Constants;
```
sam-roberts-bjss: Any reason why the NHSD prefix is missing from the using statement?

src/NHSD.BuyingCatalogue.Identity.Api/Services/ProfileService.cs
```diff
@@ -21,7 +21,13 @@ public sealed class ProfileService : IProfileService
         private static readonly IDictionary<string, IEnumerable<Claim>> _organisationFunctionClaims =
             new Dictionary<string, IEnumerable<Claim>>
             {
-                { "Authority", new List<Claim> { new Claim(ApplicationClaimTypes.Organisation, "Manage") } }
+                { 
+                    "Authority", new List<Claim>
+                    {
+                        new Claim(ApplicationClaimTypes.Organisation, ApplicationPermissions.Manage),
```
sam-roberts-bjss: Just had an idea but I wonder if it would be better to create the claims in a separate class rather than exposing all the different constants? That way the claim types and application permissions can be internal to the 'ApplicationClaims' class.

src/NHSD.BuyingCatalogue.Identity.Api/Settings/IdentityResourceSetting.cs
CallumBradburyBJSS: Yeah I missed it in my initial checkin, it's added now

src/NHSD.BuyingCatalogue.Identity.Api/Services/ProfileService.cs
```diff
@@ -3,14 +3,15 @@
 using System.IdentityModel.Tokens.Jwt;
 using System.Security.Claims;
 using System.Threading.Tasks;
+using BuyingCatalogue.Identity.Common.Constants;
```
sam-roberts-bjss: NHSD is still missing from this using statement.

src/NHSD.BuyingCatalogue.Organisations.Api/Startup.cs
```diff
@@ -57,7 +59,17 @@ public void ConfigureServices(IServiceCollection services)
 
             services.AddAuthorization(options =>
             {
-                options.AddPolicy(Policy.CanAccessOrganisation, policy => policy.RequireClaim("Organisation"));
+                options.AddPolicy(Policy.CanAccessOrganisation, policy => policy.RequireClaim(ApplicationClaimTypes.Organisation));
```
sam-roberts-bjss: I can only see this getting bigger. It might be a good idea to create a Service Collection Extensions class and add an AddCustomAuthorization method. Make a judgement call on this one. Doesn't have to be as part of the task but definitely something we need to keep an eye on.

src/NHSD.BuyingCatalogue.Identity.Api/Services/ProfileService.cs
CallumBradburyBJSS: I like the idea, but I'm inclined to make this a tech debt ticket, as it would be neater, certainly, but is a bit of work to change (mainly due to the tests) and it's currently functional and tidy.

src/NHSD.BuyingCatalogue.Organisations.Api/Startup.cs
CallumBradburyBJSS: I agree it's going to possibly get too big. It'd be nice if we had some sort of... dare I say it, PolicyBuilder (not really a builder, but something to create the policies with the appropriate claims in a nice fashion), to trim it down a bit. I wasn't really a fan of the service collection extensions, but if we can find ways to tidy up the start up class I'm all for it, if they end up being ServiceCollectionExtensions I'm not opposed to it

src/NHSD.BuyingCatalogue.Identity.Api/Services/ProfileService.cs
sam-roberts-bjss: 👍

src/NHSD.BuyingCatalogue.Organisations.Api/Startup.cs
sam-roberts-bjss: Haha! PolicyBuilder maybe an option but tidy up definitely sounds good.

PR #86, opened 2020-03-25
5447: Identity API running as /Identity

RuthBerry96:


src/NHSD.BuyingCatalogue.Identity.Api/Startup.cs
```diff
@@ -58,13 +58,17 @@ public void ConfigureServices(IServiceCollection services)
...
+            var urlPrefix = _configuration.GetValue<string>("urlPrefix");
+
             services.AddIdentityServer(options =>
                 {
                     options.Events.RaiseFailureEvents = true;
                     options.Events.RaiseErrorEvents = true;
                     options.Events.RaiseSuccessEvents = true;
                     options.IssuerUri = issuerUrl;
-                    options.UserInteraction.ErrorUrl = "/Error";
+                    options.UserInteraction.LoginUrl = $"/{urlPrefix}/account/login";
```
sam-roberts-bjss: Consistent casing? Just comparing Account to Error. 

Also if someone configures URL Prefix to an empty string or NULL to remove the prefix, would the application still work? I am not sure it would cope with `//account/login`. Look up PathString it may help in this scenario to chain the segments of a path.

src/NHSD.BuyingCatalogue.Identity.Api/Startup.cs
```diff
@@ -101,10 +107,13 @@ public void Configure(IApplicationBuilder app)
...
             else
             {
-                app.UseExceptionHandler("/Account/Error");
+                app.UseExceptionHandler($"/{urlPrefix}/Error");
             }
 
-            app.UseStaticFiles();
+            app.UseStaticFiles(new StaticFileOptions
+            {
+                RequestPath = "/identity"
```
sam-roberts-bjss: Shouldn't this be using the urlPrefix?

src/NHSD.BuyingCatalogue.Identity.Api/Startup.cs
```diff
@@ -101,10 +107,13 @@ public void Configure(IApplicationBuilder app)
             }
             else
             {
-                app.UseExceptionHandler("/Account/Error");
+                app.UseExceptionHandler($"/{urlPrefix}/Error");
```
sam-roberts-bjss: Any reason for removing the /Account from the error handler path?

src/NHSD.BuyingCatalogue.Identity.Api/Views/Shared/_Layout.cshtml
```diff
@@ -1,4 +1,10 @@
-﻿<!DOCTYPE html>
+﻿@using Microsoft.Extensions.Configuration
+@inject IConfiguration config 
+@{
+    var urlPrefix = config.GetValue<string>("urlPrefix");
```
sam-roberts-bjss: The use of the configuration key `urlPrefix` is referenced in quite a few places. Would it be possible to abstract this configuration into an `ApplicationSettings` class or something? So that the configuration property only has to be changed in one place.

RuthBerry96:
This change set will likely break some of the helm parts, as the routing on identity will be changed to  /identity. All parts of identity will have this prefixed (including the well-known.)

This was discussed with @stuart-bass-bjss

PR #89, opened 2020-03-26
Add send of e-mail following creation of new user

bjss-chris:
AB#5208

### DIFF
--- a/src/NHSD.BuyingCatalogue.Identity.Api/appsettings.Development.json
+++ b/src/NHSD.BuyingCatalogue.Identity.Api/appsettings.Development.json
@@ -9,7 +9,7 @@
       "requireClientSecret": true,
       "secret": "SampleClientSecret",
       "requireConsent": false,
-      "redirectUrls": [ "http://localhost:49176/signin-oidc" ],
+      "redirectUrls": [ "http://localhost:49176/signin-oidc", "https://host.docker.internal/admin/oauth/callback" ],
       "postLogoutRedirectUrls": [ "http://localhost:49176/signout-callback-oidc" ],
       "allowedScopes": [ "openid", "email", "profile", "SampleResource", "Organisation" ]
     },
--- a/src/NHSD.BuyingCatalogue.Identity.Api/appsettings.json
+++ b/src/NHSD.BuyingCatalogue.Identity.Api/appsettings.json
@@ -39,7 +39,7 @@
     }
   ],
   "ConnectionStrings": {
-    "CatalogueUsers": "Data Source=localhost,1451;Initial Catalog=CatalogueUsers;MultipleActiveResultSets=true;User ID=NHSD;Password=DisruptTheMarket1!"
+    "CatalogueUsers": "Data Source=localhost,1533;Initial Catalog=CatalogueUsers;MultipleActiveResultSets=true;User ID=NHSD;Password=DisruptTheMarket1!"
   },
   "cookieExpiration": {
     "expireTimeSpan": "08:00:00",
--- a/src/NHSD.BuyingCatalogue.Organisations.Api/Controllers/UsersController.cs
+++ b/src/NHSD.BuyingCatalogue.Organisations.Api/Controllers/UsersController.cs
@@ -5,6 +5,7 @@ using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc;
 using NHSD.BuyingCatalogue.Organisations.Api.Models;
 using NHSD.BuyingCatalogue.Organisations.Api.Repositories;
+using NHSD.BuyingCatalogue.Organisations.Api.Services;
 using NHSD.BuyingCatalogue.Organisations.Api.ViewModels.Users;
 
 namespace NHSD.BuyingCatalogue.Organisations.Api.Controllers
@@ -15,11 +16,15 @@ namespace NHSD.BuyingCatalogue.Organisations.Api.Controllers
     [Produces("application/json")]
     public sealed class UsersController : Controller
     {
+        private readonly IRegistrationService _registrationService;
         private readonly IUsersRepository _usersRepository;
 
-        public UsersController(IUsersRepository usersRepository)
+        public UsersController(
+            IUsersRepository usersRepository,
+            IRegistrationService registrationService)
         {
             _usersRepository = usersRepository ?? throw new ArgumentNullException(nameof(usersRepository));
+            _registrationService = registrationService ?? throw new ArgumentNullException(nameof(registrationService));
         }
 
         [HttpGet]
@@ -63,8 +68,13 @@ namespace NHSD.BuyingCatalogue.Organisations.Api.Controllers
                 OrganisationFunction = "Buyer",
                 CatalogueAgreementSigned = false
             };
-            
+
             await _usersRepository.CreateUserAsync(newApplicationUser);
+
+            // TODO: discuss exception handling options 
+            // TODO: consider moving sending e-mail out of process
+            // (the current in-process implementation has a significant impact on response time)
+            await _registrationService.SendInitialEmailAsync(newApplicationUser);
 
             return Ok();
         }
--- a/src/NHSD.BuyingCatalogue.Organisations.Api/Models/ApplicationUser.cs
+++ b/src/NHSD.BuyingCatalogue.Organisations.Api/Models/ApplicationUser.cs
@@ -9,7 +9,7 @@ namespace NHSD.BuyingCatalogue.Organisations.Api.Models
 
         public string LastName { get; set; }
 
-        public string DisplayName => FirstName + " " + LastName;
+        public string DisplayName => $"{FirstName} {LastName}";
 
         public Guid PrimaryOrganisationId { get; set; }
 
--- a/src/NHSD.BuyingCatalogue.Organisations.Api/Models/ApplicationUser.cs
+++ b/src/NHSD.BuyingCatalogue.Organisations.Api/Models/ApplicationUser.cs
@@ -9,6 +9,8 @@ namespace NHSD.BuyingCatalogue.Organisations.Api.Models
 
         public string LastName { get; set; }
 
+        public string DisplayName => FirstName + " " + LastName;
+
         public Guid PrimaryOrganisationId { get; set; }
 
         public string OrganisationFunction { get; set; }
--- a/src/NHSD.BuyingCatalogue.Organisations.Api/Services/EmailAddress.cs
+++ b/src/NHSD.BuyingCatalogue.Organisations.Api/Services/EmailAddress.cs
@@ -1,5 +1,4 @@
 ﻿using System;
-using MimeKit;
 
 namespace NHSD.BuyingCatalogue.Organisations.Api.Services
 {
@@ -56,11 +55,5 @@ namespace NHSD.BuyingCatalogue.Organisations.Api.Services
                 _address = value;
             }
         }
-
-        /// <summary>
-        /// Returns the receiver as a <see cref="MailboxAddress"/>.
-        /// </summary>
-        /// <returns>the corresponding <see cref="MailboxAddress"/>.</returns>
-        internal MailboxAddress AsMailboxAddress() => new MailboxAddress(DisplayName, Address);
     }
 }
--- a/src/NHSD.BuyingCatalogue.Organisations.Api/Services/EmailAddress.cs
+++ b/src/NHSD.BuyingCatalogue.Organisations.Api/Services/EmailAddress.cs
@@ -1,0 +1,66 @@
+﻿using System;
+using MimeKit;
+
+namespace NHSD.BuyingCatalogue.Organisations.Api.Services
+{
+    /// <summary>
+    /// An e-mail address.
+    /// </summary>
+    internal sealed class EmailAddress
+    {
+        private string address;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="EmailAddress"/> class.
+        /// </summary>
+        /// <remarks>Required by <see cref="System.Text.Json"/>.</remarks>
+        public EmailAddress()
+        {
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="EmailAddress"/> class
+        /// with the given name and address.
+        /// </summary>
+        /// <param name="name">An optional </param>
+        /// <param name="address"></param>
+        internal EmailAddress(string name, string address)
+        {
+            Name = name;
+            Address = address;
+        }
+
+        /// <summary>
+        /// Gets or sets the display name of the address.
+        /// </summary>
+        /// <remarks>An optional display name, for example
+        /// Buying Catalogue Team.</remarks>
+        public string Name { get; set; }
+
+        /// <summary>
+        /// Gets or sets the actual e-mail address.
+        /// </summary>
+        /// <exception cref="ArgumentNullException"><paramref name="value"/> is <see langref="null"/>.</exception>
+        /// <exception cref="ArgumentException"><paramref name="value"/> is empty or white space.</exception>
+        public string Address
+        {
+            get => address;
+            set
+            {
+                if (value == null)
+                    throw new ArgumentNullException(nameof(value));
+
+                if (string.IsNullOrWhiteSpace(value))
+                    throw new ArgumentException(nameof(value));
+
+                address = value;
+            }
+        }
+
+        /// <summary>
+        /// Returns the receiver as a <see cref="MailboxAddress"/>.
+        /// </summary>
+        /// <returns>the corresponding <see cref="MailboxAddress"/>.</returns>
+        internal MailboxAddress AsMailboxAddress() => new MailboxAddress(Name, Address);
+    }
+}
--- a/src/NHSD.BuyingCatalogue.Organisations.Api/Services/EmailAddress.cs
+++ b/src/NHSD.BuyingCatalogue.Organisations.Api/Services/EmailAddress.cs
@@ -51,7 +51,7 @@ namespace NHSD.BuyingCatalogue.Organisations.Api.Services
                     throw new ArgumentNullException(nameof(value));
 
                 if (string.IsNullOrWhiteSpace(value))
-                    throw new ArgumentException(nameof(value));
+                    throw new ArgumentException($"{nameof(Address)} is required.", nameof(value));
 
                 address = value;
             }
--- a/src/NHSD.BuyingCatalogue.Organisations.Api/Services/EmailAddress.cs
+++ b/src/NHSD.BuyingCatalogue.Organisations.Api/Services/EmailAddress.cs
@@ -51,7 +51,7 @@ namespace NHSD.BuyingCatalogue.Organisations.Api.Services
                     throw new ArgumentNullException(nameof(value));
 
                 if (string.IsNullOrWhiteSpace(value))
-                    throw new ArgumentException($"{nameof(Address)} is required.", nameof(value));
+                    throw new ArgumentException($"{nameof(value)} cannot be empty or white space.", nameof(value));
 
                 address = value;
             }
--- a/src/NHSD.BuyingCatalogue.Organisations.Api/Services/EmailAddress.cs
+++ b/src/NHSD.BuyingCatalogue.Organisations.Api/Services/EmailAddress.cs
@@ -20,13 +20,13 @@ namespace NHSD.BuyingCatalogue.Organisations.Api.Services
 
         /// <summary>
         /// Initializes a new instance of the <see cref="EmailAddress"/> class
-        /// with the given name and address.
+        /// with the given display name and address.
         /// </summary>
-        /// <param name="name">An optional </param>
+        /// <param name="displayName">An optional display name.</param>
         /// <param name="address"></param>
-        internal EmailAddress(string name, string address)
+        internal EmailAddress(string displayName, string address)
         {
-            Name = name;
+            DisplayName = displayName;
             Address = address;
         }
 
@@ -35,7 +35,7 @@ namespace NHSD.BuyingCatalogue.Organisations.Api.Services
         /// </summary>
         /// <remarks>An optional display name, for example
         /// Buying Catalogue Team.</remarks>
-        public string Name { get; set; }
+        public string DisplayName { get; set; }
 
         /// <summary>
         /// Gets or sets the actual e-mail address.
@@ -61,6 +61,6 @@ namespace NHSD.BuyingCatalogue.Organisations.Api.Services
         /// Returns the receiver as a <see cref="MailboxAddress"/>.
         /// </summary>
         /// <returns>the corresponding <see cref="MailboxAddress"/>.</returns>
-        internal MailboxAddress AsMailboxAddress() => new MailboxAddress(Name, Address);
+        internal MailboxAddress AsMailboxAddress() => new MailboxAddress(DisplayName, Address);
     }
 }
--- a/src/NHSD.BuyingCatalogue.Organisations.Api/Services/EmailAddress.cs
+++ b/src/NHSD.BuyingCatalogue.Organisations.Api/Services/EmailAddress.cs
@@ -8,7 +8,7 @@ namespace NHSD.BuyingCatalogue.Organisations.Api.Services
     /// </summary>
     internal sealed class EmailAddress
     {
-        private string address;
+        private string _address;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="EmailAddress"/> class.
@@ -44,7 +44,7 @@ namespace NHSD.BuyingCatalogue.Organisations.Api.Services
         /// <exception cref="ArgumentException"><paramref name="value"/> is empty or white space.</exception>
         public string Address
         {
-            get => address;
+            get => _address;
             set
             {
                 if (value == null)
@@ -53,7 +53,7 @@ namespace NHSD.BuyingCatalogue.Organisations.Api.Services
                 if (string.IsNullOrWhiteSpace(value))
                     throw new ArgumentException($"{nameof(value)} cannot be empty or white space.", nameof(value));
 
-                address = value;
+                _address = value;
             }
         }
 
--- a/src/NHSD.BuyingCatalogue.Organisations.Api/Services/EmailMessage.cs
+++ b/src/NHSD.BuyingCatalogue.Organisations.Api/Services/EmailMessage.cs
@@ -1,0 +1,108 @@
+﻿using System;
+using MimeKit;
+
+namespace NHSD.BuyingCatalogue.Organisations.Api.Services
+{
+    /// <summary>
+    /// An e-mail message.
+    /// </summary>
+    internal sealed class EmailMessage
+    {
+        /// <summary>
+        /// The placeholder to use in any body text that will be replaced by the URL
+        /// for the password reset endpoint.
+        /// </summary>
+        internal const string ResetPasswordLinkPlaceholder = "[ResetPasswordLink]";
+
+        private EmailAddress sender;
+        private EmailAddress recipient;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="EmailMessage"/> class.
+        /// </summary>
+        /// <remarks>Required by <see cref="System.Text.Json"/>.</remarks>.
+        public EmailMessage()
+        {
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="EmailMessage"/> class based
+        /// on the supplied <paramref name="message"/>.
+        /// </summary>
+        /// <param name="message">The message to use as the basis of the newly constructed message.</param>
+        /// <param name="passwordResetUrl">The password reset endpoint.</param>
+        /// <remarks>This constructor will initialize a new <see cref="EmailMessage"/> using
+        /// the sender, subject and bodies of the <paramref name="message"/> parameter.
+        /// Any password reset placeholders (<see cref="ResetPasswordLinkPlaceholder"/> will
+        /// be replaced by the value of <paramref name="passwordResetUrl"/>.</remarks>
+        internal EmailMessage(EmailMessage message, Uri passwordResetUrl)
+        {
+            Sender = message.Sender;
+            Subject = message.Subject;
+
+            var passwordResetLink = passwordResetUrl.ToString();
+
+            HtmlBody = message.HtmlBody.Replace(ResetPasswordLinkPlaceholder, passwordResetLink);
+            TextBody = message.TextBody.Replace(ResetPasswordLinkPlaceholder, passwordResetLink);
+        }
+
+        /// <summary>
+        /// Gets or sets the sender (from address) of the message.
+        /// </summary>
+        /// <exception cref="ArgumentNullException"><paramref name="value"/> is <see langref="null"/>.</exception>
+        public EmailAddress Sender
+        {
+            get => sender;
+            set => sender = value ?? throw new ArgumentNullException(nameof(value));
+        }
+
+        /// <summary>
+        /// Gets or sets the subject of the message.
+        /// </summary>
+        public string Subject { get; set; }
+
+        /// <summary>
+        /// Gets or sets the HTML version of the body.
+        /// </summary>
+        public string HtmlBody { get; set; }
+
+        /// <summary>
+        /// Gets or sets the plain text version of the body.
+        /// </summary>
+        public string TextBody { get; set; }
+
+        /// <summary>
+        /// Gets or sets the recipient (to address) of the message.
+        /// </summary>
+        /// /// <exception cref="ArgumentNullException"><paramref name="value"/> is <see langref="null"/>.</exception>
+        internal EmailAddress Recipient
+        {
+            get => recipient;
+            set => recipient = value ?? throw new ArgumentNullException(nameof(value));
+        }
+
+        /// <summary>
+        /// Returns the receiver as a <see cref="MimeMessage"/>.
+        /// </summary>
+        /// <returns>the corresponding <see cref="MimeMessage"/>.</returns>
+        internal MimeMessage AsMimeMessage()
+        {
+            var bodyBuilder = new BodyBuilder
+            {
+                HtmlBody = HtmlBody,
+                TextBody = TextBody,
+            };
+
+            var message = new MimeMessage
+            {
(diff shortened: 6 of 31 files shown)


src/NHSD.BuyingCatalogue.Identity.Api/appsettings.json
```diff
@@ -39,7 +39,7 @@
     }
   ],
   "ConnectionStrings": {
-    "CatalogueUsers": "Data Source=localhost,1451;Initial Catalog=CatalogueUsers;MultipleActiveResultSets=true;User ID=NHSD;Password=DisruptTheMarket1!"
+    "CatalogueUsers": "Data Source=localhost,1533;Initial Catalog=CatalogueUsers;MultipleActiveResultSets=true;User ID=NHSD;Password=DisruptTheMarket1!"
```
assoomro: Please revert this to 1451 as having it on 1533 will throw off people who're working with the local k8s cluster.
If you find it annoying having to overwrite this to use the docker db, I suggest you use [User Secrets](https://docs.microsoft.com/en-us/aspnet/core/security/app-secrets?view=aspnetcore-3.1&tabs=windows).

src/NHSD.BuyingCatalogue.Organisations.Api/Models/ApplicationUser.cs
```diff
@@ -9,6 +9,8 @@ public sealed class ApplicationUser : IdentityUser
 
         public string LastName { get; set; }
 
+        public string DisplayName => FirstName + " " + LastName;
```
assoomro: Just a suggestion as it boils down to personal preference. Have you considered using string interpolation in this case?

bjss-chris: I could've done but chose not to.

`public string DisplayName => $"{FirstName} {LastName}";`

Maybe the string interpolation looks slightly neater... I'm not sure. The string concatenation version just seems to read ever so slightly better to me. However, I'm easy on this one.

src/NHSD.BuyingCatalogue.Organisations.Api/Services/MailKitEmailService.cs
```diff
@@ -0,0 +1,52 @@
...
+        {
+            _client = client;
+
+            // TODO: confirm this is acceptable for us if using SSL
+            // (as we will be using a trusted server that is not
+            // user-configurable this should be fine, providing the
+            // root CA is trusted by the host system)
+            client.ServerCertificateValidationCallback = (s, c, h, e) => true;
+
+            _settings = settings;
```
assoomro: Can't confirm this one unfortunately. I'll ask either @sam-roberts-bjss @CallumBradburyBJSS for a second pair of eyes.

src/NHSD.BuyingCatalogue.Identity.Api/appsettings.json
bjss-chris: Why do we have so many different port numbers flying around? Why is it different between Kubernetes and docker and why is this a problem here but not in the organisations API?

src/NHSD.BuyingCatalogue.Organisations.Api/Services/RegistrationService.cs
```diff
@@ -0,0 +1,44 @@
...
+        }
+
+        /// <summary>
+        /// Asynchronously sends the initial e-mail to the specified <paramref name="user"/>.
+        /// </summary>
+        /// <param name="user">The user to send the e-mail to.</param>
+        /// <returns>An asynchronous task context.</returns>
+        public async Task SendInitialEmailAsync(ApplicationUser user)
+        {
+            // TODO: replace hard-coded Uri with reset password endpoint once available
```
assoomro: Have we added task/s for replacing these values yet? If not can we please make them?

src/NHSD.BuyingCatalogue.Organisations.Api/appsettings.json
```diff
@@ -10,8 +10,26 @@
...
+      "TextBody": "Thank you for requesting a user account for the Buying Catalogue.\n\nYour account has now been created and you can use this link to set your password.\n\n[ResetPasswordLink]\n\nYour password must be at least ten characters in length and match on confirmation.\n\nIt must contain at least 3 of the following 4 character types:\n\nlower case letters (a-z)\nupper case letters (A-Z)\nnumbers (0-9)\nspecial characters (e.g. !@#$%^&*)\n\nThe link can be used only once to set up one account, and will be valid for 24 hours from the time you received this email.\n\nYours sincerely,\nThe Buying Catalogue team"
```
assoomro: I wonder whether this is the optimal way to set the content of the email. Would it perhaps be better to obtain the content through a reading a file?

tests/NHSD.BuyingCatalogue.Organisations.Api.UnitTests/Services/EmailAddressTests.cs
```diff
@@ -0,0 +1,56 @@
...
+using NUnit.Framework;
+
+namespace NHSD.BuyingCatalogue.Organisations.Api.UnitTests.Services
+{
+    [TestFixture]
+    [Parallelizable(ParallelScope.All)]
+    internal sealed class EmailAddressTests
+    {
+        [Test]
+        public void Constructor_String_String_InitializesExpectedValues()
```
assoomro: I think this might be a typo.

src/NHSD.BuyingCatalogue.Organisations.Api/Models/ApplicationUser.cs
CallumBradburyBJSS: I prefer the $ one, for sure

src/NHSD.BuyingCatalogue.Organisations.Api/Controllers/UsersController.cs
```diff
@@ -15,11 +16,15 @@ namespace NHSD.BuyingCatalogue.Organisations.Api.Controllers
...
+        private readonly IRegistrationService _registrationService;
         private readonly IUsersRepository _usersRepository;
 
-        public UsersController(IUsersRepository usersRepository)
+        public UsersController(
+            IUsersRepository usersRepository,
+            IRegistrationService registrationService)
         {
             _usersRepository = usersRepository ?? throw new ArgumentNullException(nameof(usersRepository));
+            _registrationService = registrationService ?? throw new ArgumentNullException(nameof(registrationService));
```
CallumBradburyBJSS: We have a cool .ThrowIfNull() extension that is sporadically used, and could maybe be used here

src/NHSD.BuyingCatalogue.Organisations.Api/Services/MailKitEmailService.cs
```diff
@@ -0,0 +1,52 @@
...
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="MailKitEmailService"/> class
+        /// using the provided <paramref name="client"/> and <paramref name="settings"/>.
+        /// </summary>
+        /// <param name="client">The mail transport to use to send e-mail.</param>
+        /// <param name="settings">The SMTP configuration.</param>
+        public MailKitEmailService(IMailTransport client, SmtpSettings settings)
+        {
+            _client = client;
```
CallumBradburyBJSS: worth checking for null here as we do it everywhere else?

tests/NHSD.BuyingCatalogue.Organisations.Api.UnitTests/Services/EmailAddressTests.cs
bjss-chris: What's that – `String_String`? That's actually deliberate – it corresponds to the signature of the constructor.

src/NHSD.BuyingCatalogue.Organisations.Api/Services/RegistrationService.cs
bjss-chris: Good point.

src/NHSD.BuyingCatalogue.Organisations.Api/Services/MailKitEmailService.cs
CallumBradburyBJSS: The amount of clues I have on this one is zero

src/NHSD.BuyingCatalogue.Organisations.Api/Controllers/UsersController.cs
```diff
@@ -63,9 +68,14 @@ public async Task<ActionResult> CreateUserAsync(Guid organisationId, CreateUserR
                 OrganisationFunction = "Buyer",
                 CatalogueAgreementSigned = false
             };
-            
+
             await _usersRepository.CreateUserAsync(newApplicationUser);
 
+            // TODO: discuss exception handling options 
+            // TODO: consider moving sending e-mail out of process
+            // (the current in-process implementation has a significant impact on response time)
```
sam-roberts-bjss: I agree with the comments but it might be more of a style thing but I would rather raise a task on DevOps than leave the TODO comment within the application. I have seen TODO comments just get ignored & left in an application for years (You could say this about some DevOps tasks). At least more roles within the team would see we are introducing tech debt by doing it this way.

I mentioned it to @michaelpace-bjss so he is aware of it. Agreed that for now we will leave it how you've implemented it, but will need to change it soon.

src/NHSD.BuyingCatalogue.Organisations.Api/Services/EmailAddress.cs
```diff
@@ -0,0 +1,66 @@
...
+        /// Gets or sets the actual e-mail address.
+        /// </summary>
+        /// <exception cref="ArgumentNullException"><paramref name="value"/> is <see langref="null"/>.</exception>
+        /// <exception cref="ArgumentException"><paramref name="value"/> is empty or white space.</exception>
+        public string Address
+        {
+            get => address;
+            set
+            {
+                if (value == null)
```
sam-roberts-bjss: Any reason for the two statements here? `String.IsNullOrEmpty()` will carry out the first conditional check.

src/NHSD.BuyingCatalogue.Organisations.Api/Services/EmailAddress.cs
```diff
@@ -0,0 +1,66 @@
...
+        {
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="EmailAddress"/> class
+        /// with the given name and address.
+        /// </summary>
+        /// <param name="name">An optional </param>
+        /// <param name="address"></param>
+        internal EmailAddress(string name, string address)
```
sam-roberts-bjss: Would it be better to make this class immutable? I was under the impression that throwing an exception on the setting of the Address property was an anti pattern. It would not be obvious to a developer using this class that it throws. To reduce the surface area to get something wrong, would it be better to encapsulate this & not expose a setter? This obviously depends on how the object is used throughout the rest of the application.

src/NHSD.BuyingCatalogue.Organisations.Api/Services/MailKitEmailService.cs
```diff
@@ -0,0 +1,52 @@
...
+            _settings = settings;
+        }
+
+        /// <summary>
+        /// Sends an e-mail message asynchronously using the
+        /// provided <see cref="IMailTransport"/> instance.
+        /// </summary>
+        /// <param name="emailMessage">The e-mail message to send asynchronously.</param>
+        /// <returns>An asynchronous task context.</returns>
+        public async Task SendEmailAsync(EmailMessage emailMessage)
```
sam-roberts-bjss: Could or should the email address parameter be null?

src/NHSD.BuyingCatalogue.Organisations.Api/Services/MailKitEmailService.cs
bjss-chris: I guess there are two questions here:

1. Will SSL be required to connect to the SMTP server in any environment?
2. If yes, will the root CA be trusted by the host OS by default?

If the answer is 1 = Yes, 2 = No then we will need to update the callback to handle an `SslHandshakeException` otherwise we don't need to do anything (other than remove the comment).

(We wouldn't need to make any changes to work with TLS. I've tested using a TLS-enabled server and it works as is.)

Maybe one for @stuart-bass-bjss?

src/NHSD.BuyingCatalogue.Organisations.Api/Services/MailKitEmailService.cs
```diff
@@ -0,0 +1,52 @@
...
+        {
+            await _client.ConnectAsync(_settings.Host, _settings.Port, _settings.UseSsl);
```
sam-roberts-bjss: Is IMailTransport disposable? I ask as I think it would mean this service would have to be disposable as well.

src/NHSD.BuyingCatalogue.Organisations.Api/Services/EmailAddress.cs
```diff
@@ -0,0 +1,66 @@
...
+            Name = name;
+            Address = address;
+        }
+
+        /// <summary>
+        /// Gets or sets the display name of the address.
+        /// </summary>
+        /// <remarks>An optional display name, for example
+        /// Buying Catalogue Team.</remarks>
+        public string Name { get; set; }
```
sam-roberts-bjss: Could this be named DisplayName? Initially I thought this was the name of the user.

src/NHSD.BuyingCatalogue.Organisations.Api/Settings/RegistrationSettings.cs
```diff
@@ -0,0 +1,16 @@
+﻿using NHSD.BuyingCatalogue.Organisations.Api.Services;
+
+namespace NHSD.BuyingCatalogue.Organisations.Api.Settings
+{
+    /// <summary>
+    /// Registration settings.
+    /// </summary>
+    internal sealed class RegistrationSettings
```
sam-roberts-bjss: Do we need a `RegistrationSettings` class if it only contains the email message. Could EmailMessage be a parameter on the RegistrationService instead? It would reduce the amount of code & we don't know yet whether we will need anymore settings, unless you have some insight?

src/NHSD.BuyingCatalogue.Organisations.Api/Settings/SmtpAuthenticationSettings.cs
```diff
@@ -0,0 +1,23 @@
+﻿namespace NHSD.BuyingCatalogue.Organisations.Api.Settings
+{
+    /// <summary>
+    /// SMTP authentication settings.
+    /// </summary>
+    internal sealed class SmtpAuthenticationSettings
```
sam-roberts-bjss: Could this class be made immutable? It will make it clearer what properties are required if any.

src/NHSD.BuyingCatalogue.Organisations.Api/Settings/SmtpSettings.cs
```diff
@@ -0,0 +1,29 @@
+﻿namespace NHSD.BuyingCatalogue.Organisations.Api.Settings
+{
+    /// <summary>
+    /// SMTP server settings.
+    /// </summary>
+    internal sealed class SmtpSettings
```
sam-roberts-bjss: Could the class `SmtpSettings` class be made immutable? Once you have created an instance of this class do the values want to change again?

src/NHSD.BuyingCatalogue.Organisations.Api/Services/MailKitEmailService.cs
stuart-bass-bjss: please push into config. Prod will be SSL, and hopefully trusted, but it's likely local/dev/test will be ssl but self signed. Please use the existing AllowInvalidCertificate config value to set whether invalid ssl certificates are set. Startup already does this for calls to isapi. The recomendation is to set HttpClientHandler.DangerousAcceptAnyServerCertificateValidator as the http handler if you can for this.

src/NHSD.BuyingCatalogue.Identity.Api/appsettings.json
stuart-bass-bjss: Please can you add as discussed two sets of launch settings to both isapi & opai, one for running against dev (k8s cluster), and one for running against the integration environment (like bapi is set up).

tests/NHSD.BuyingCatalogue.Organisations.Api.UnitTests/Services/EmailAddressTests.cs
assoomro: 😅 Ignore what I said

src/NHSD.BuyingCatalogue.Organisations.Api/Controllers/UsersController.cs
bjss-chris: I'm not a fan of `.ThrowIfNull()`.

While it is obviously syntactically fine semantically it is wrong to me – it shouldn't be possible to invoke a method on a null object – although we can due to the way extension methods are implemented.

I also think extension methods that essentially extend everything just smell – Mr Lippert had some [comments](https://stackoverflow.com/a/7652359) on this a few years back.

This approach also means that the call stack shows the exception occurring in the correct method rather than an extension method – admittedly this is a small difference in this case but another reason I prefer the version above.

Perhaps another one for the next code standards meeting...

src/NHSD.BuyingCatalogue.Organisations.Api/Models/ApplicationUser.cs
bjss-chris: That's two votes for the string interpolation option so I'll switch it to that.

tests/NHSD.BuyingCatalogue.Organisations.Api.UnitTests/Services/EmailAddressTests.cs
bjss-chris: Good to check though!

src/NHSD.BuyingCatalogue.Organisations.Api/Services/MailKitEmailService.cs
bjss-chris: I could do with testing this. Any suggestions for a quick way to spin up an SMTP server with SSL?

@assoomro do you know if MailDev supports SSL?

src/NHSD.BuyingCatalogue.Organisations.Api/Services/RegistrationService.cs
bjss-chris: Tasks 5718 and 5719 created against story 3538 (create user) for discussing exception handling options and adding SMTP health checks.

Task 5720 created against the tech debt story (5648) to move e-mailing out of process.

We can obviously reassign these tasks a different parent (e.g. tech debt story) if we think that might be more appropriate.

src/NHSD.BuyingCatalogue.Organisations.Api/Services/MailKitEmailService.cs
assoomro: @bjss-chris not from the top of my head, I'll do some digging and come back to you :]

assoomro: @bjss-chris https://github.com/maildev/maildev#usage there seems to be a flag to enable ssl for outgoing email.
We can raise a task to add this to the config (local docker-compose)

bjss-chris: @assoomro As that's outgoing I suspect that is for use where you have MailDev set up to relay to messages whereas what we need here is to be able to connect to MailDev using SSL (incoming).

assoomro: You're quite right, Did some more digging and they haven't put that in yet. I'm afraid we won't be able to test this then. You reckon it might be worth to switch over to a SMTP server that provides SSL?

assoomro: To add to my previous response - we can switch to Smtp4Dev which [supports ssl] (https://github.com/smorks/smtp4dev/blob/master/Rnwood.SmtpServer/DefaultServer.cs) Given that we've everything configurable through appsettings.json, it shouldn't take too long to change.

bjss-chris: I think I'll have a look to see how easy/difficult it is to spin up locally in the meantime so I can start doing some testing.

bjss-chris: @michaelpace-bjss has confirmed we do need to support TLS 1.2 but not SSL.

The existing implementation already supports TLS and @assoomro has a PR to switch from MailDev to Smtp4dev so this should be fine as is (but I'll remove the TODO comments).

bjss-chris: Although thinking about it some more I think we do still need to make a tweak to be able to handle cases such as @stuart-bass-bjss mentioned where we will be using self-signed certs – I'll try running some tests...

bjss-chris: It looks like Smtp4dev defaults to TLS 1.0 so we'll either need to check if there's any configuration we can set or look for an alternative SMTP server for our integration tests.

src/NHSD.BuyingCatalogue.Organisations.Api/Services/MailKitEmailService.cs
bjss-chris: I'm not sure we need to here.

The class is `internal` so (with the exception of tests) it should always be constructed by the DI container. If the `IMailTransport` dependency hasn't been wired up correctly then the DI container will throw an exception.

src/NHSD.BuyingCatalogue.Organisations.Api/Controllers/UsersController.cs
bjss-chris: I don't think there's any harm in having the TODO comments given that a change is always reviewed by at least one person. While I agree that there should be a corresponding task having the TODO comment means that it should prompt discussions. It's also a reminder to add a task (or check that a task has been added) should I (or someone else) forget to add one.

If the TODO comments are still there the next time someone else works on the code, hopefully it will prompt them to ask: does this still need doing? I also like that they appear in the task list in VS although I'm not sure how many people are familiar with that feature.

I'll leave them in for now but we can discuss at our next code standards meeting.

The corresponding tasks in this case are 5718 and 5720.

src/NHSD.BuyingCatalogue.Organisations.Api/Services/EmailAddress.cs
bjss-chris: There are two separate statements so that the most specific exception type can be thrown: `ArgumentNullException` in the null case and `ArgumentException` in the other cases.

While `string.IsNullOrWhiteSpace` and `string.IsNullOrEmpty` obviously both check for a null string we'd end up throwing `ArgumentException` in every case. While it's a small difference, I think it's preferable to throw the most explicit exception we can – it just means we end up covering the null case twice. It would be good if MS could provide some nice solution for this but I'm not aware of one.

src/NHSD.BuyingCatalogue.Organisations.Api/Services/EmailAddress.cs
bjss-chris: While it is bad practice to throw an exception from the `get` it is acceptable throw on `set` – a number of framework classes do this, for example [StringBuilder.Capacity](https://docs.microsoft.com/en-us/dotnet/api/system.text.stringbuilder.capacity?view=netframework-4.8#exceptions).

In many cases, I prefer immutable classes (and the first draft of this class was immutable) but I can't make `EmailAddress` immutable due to its use as configuration (the mutable properties are required for deserialization).

As we're not developing a public library I'm not sure we'd gain much by refactoring here. Even if we were developing a public library I think I'd still be inclined not to refactor (the `Address` set exceptions). Other than the exception occurring slightly sooner, I'm not sure how an exception thrown in a constructor would be any more obvious than one thrown in the set. Either way, it's a '[boneheaded exception](https://ericlippert.com/2008/09/10/vexing-exceptions/)' (Lippert) that's going to occur if a developer tries to set `Address` to `null`.

src/NHSD.BuyingCatalogue.Organisations.Api/Services/MailKitEmailService.cs
bjss-chris: I'm doing a reverse-ferret here because @sam-roberts-bjss asked a related question too.

While I'm in favour of failing fast I'm not always convinced by the value of precondition assertions in non-public code although I guess you could argue that's slightly contradictory... However, as both you and @sam-roberts-bjss have raised similar points I'm going to add them.

Possibly another one for code standards – do we want to do this everywhere? I guess nullable reference types might help here too (if we enabled that features).

src/NHSD.BuyingCatalogue.Organisations.Api/Services/MailKitEmailService.cs
bjss-chris: It shouldn't be null but it could null although only through developer error at this point in time.

As I've commented above, while I'm in favour of failing fast I'm not always convinced by the value of precondition assertions in non-public code although I guess you could argue that's slightly contradictory... However, as both you and @CallumBradburyBJSS  have raised similar points I'm going to add them.

Possibly another one for code standards – do we want to do this everywhere? I guess nullable reference types might help here too (if we enabled that features).

src/NHSD.BuyingCatalogue.Organisations.Api/Services/MailKitEmailService.cs
bjss-chris: `IMailTransport` is disposable but we shouldn't need to implement `IDisposable` here because `MailKitEmailService` does not own the `SmtpClient` instance – the DI container does so it is responsible for disposing of it.

One can argue having `IMailTransport` derive from `IDisposable` is a bit of a smell – as the use of unmanaged resources depends on the implementation not the abstraction. Unfortunately, it's not something we have control of in this case.

src/NHSD.BuyingCatalogue.Organisations.Api/Services/EmailAddress.cs
bjss-chris: In our case it will be the name of the user for recipients (or the name of a team in the case of our sender) or did you think it was the user's username?

I'm happy to rename this as suggested, regardless.

src/NHSD.BuyingCatalogue.Organisations.Api/Settings/SmtpAuthenticationSettings.cs
bjss-chris: Similar to `EmailMessage` I can't make this one immutable due to its use in configuration – the properties need to be mutable for deserialization.

The required properties depend on the value of `IsRequired`, i.e. `UserName` and `Password` are only required when `IsRequired == true`.

Hopefully this one should be fairly self-explanatory (in terms of what is required when). However, I will make a change so that the SMTP authentication settings can be omitted entirely when authentication is not required.

src/NHSD.BuyingCatalogue.Organisations.Api/appsettings.json
bjss-chris: I did ponder this implementation a little. My very first implementation just relied on constants but I thought it would be better to allow the message content to be configurable given how often the content has change on the FE.

While app settings isn't really intended for content management our content is relatively simple and we get well-understood configuration management out of the box. Given that the default content is provided via `appSettings.json` I'm not sure placing the content in a different file would bring any performance benefit – particularly as the content is currently only read once on startup.

However, I recognize there is room for improvement here so if you or anyone else has any suggestions please let me know.

src/NHSD.BuyingCatalogue.Organisations.Api/Services/MailKitEmailService.cs
bjss-chris: I've updated the implementation so it will allow an untrusted certificate only when the `AllowInvalidCertificate` flag is set in configuration.

There is a flag specifically for the SMTP settings but it will inherit from the app-level settings when not set explicitly.

bjss-chris: Now all that's need is to make sure test SMTP server supports TLS 1.2 but @assoomro already has that sorted. 👍

assoomro: Update: I've used stunnel as a TLS wrapper for maildev [commit here](https://github.com/nhs-digital-gp-it-futures/BuyingCatalogueIdentity/pull/89/commits/aacaf54fe2bc1ebd4fd861fd5b8f93f0e55d3cee)

src/NHSD.BuyingCatalogue.Organisations.Api/appsettings.json
sam-roberts-bjss: This is fine for now, isn't that pleasant to look at or I think that easy to manage in this form. I'd probably want a HTML file so I can design / style the email template. We do have a database to store things, again not that I am suggesting we use that, but wouldn't have to worry about backing up the data. It kind of depends what sort of functionality the client wants going forward e.g. different email templates per user type? / dynamic content. Simplest approach for now, enhance functionality at a later date.

src/NHSD.BuyingCatalogue.Organisations.Api/Services/EmailAddress.cs
sam-roberts-bjss: One suggestion would be to separate out the concerns of reading the email address settings and the immutability of the model. We will need to at some point validate all configuration at start up. Here is one resource I found, sure there is plenty more [Configuration validation on startup](https://andrewlock.net/adding-validation-to-strongly-typed-configuration-objects-in-asp-net-core/). By validating the configuration we can ensure the address property is not null & potentially provide a customized error message.

Although I haven't tried it, I believe Json.Net does support deserialization using a constructor [Deserialize using constructor](https://www.newtonsoft.com/json/help/html/DeserializeConstructorHandling.htm). It would mean bringing in another dependency into the project just to support this, which is a little drastic. One to keep an eye on if more requirements come in for this type of process.

As it stands I believe the above change will produce the same result & is the simplest. Just wanted to provide a couple more suggestions.

src/NHSD.BuyingCatalogue.Organisations.Api/Services/EmailAddress.cs
```diff
@@ -0,0 +1,66 @@
...
+using MimeKit;
+
+namespace NHSD.BuyingCatalogue.Organisations.Api.Services
+{
+    /// <summary>
+    /// An e-mail address.
+    /// </summary>
+    internal sealed class EmailAddress
+    {
+        private string address;
```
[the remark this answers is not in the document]
bjss-chris: Yes, we did but as I detest the underscore notation I never use it while I'm developing and only change it prior to committing. I've obviously missed a couple here though!

I'll 'fix' it.

src/NHSD.BuyingCatalogue.Organisations.Api/Services/EmailMessage.cs
```diff
@@ -0,0 +1,108 @@
...
+    internal sealed class EmailMessage
+    {
+        /// <summary>
+        /// The placeholder to use in any body text that will be replaced by the URL
+        /// for the password reset endpoint.
+        /// </summary>
+        internal const string ResetPasswordLinkPlaceholder = "[ResetPasswordLink]";
+
+        private EmailAddress sender;
+        private EmailAddress recipient;
```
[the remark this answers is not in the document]
bjss-chris: I'll 'fix' it as above.

src/NHSD.BuyingCatalogue.Identity.Api/appsettings.Development.json
```diff
@@ -9,7 +9,7 @@
       "requireClientSecret": true,
       "secret": "SampleClientSecret",
       "requireConsent": false,
-      "redirectUrls": [ "http://localhost:49176/signin-oidc" ],
+      "redirectUrls": [ "http://localhost:49176/signin-oidc", "https://host.docker.internal/admin/oauth/callback" ],
```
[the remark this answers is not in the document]
bjss-chris: @stuart-bass-bjss introduced this so that the admin FE works when running Identity in VS (F5-ing it)

src/NHSD.BuyingCatalogue.Organisations.Api/Services/EmailAddress.cs
```diff
@@ -0,0 +1,66 @@
...
+
+                address = value;
+            }
+        }
+
+        /// <summary>
+        /// Returns the receiver as a <see cref="MailboxAddress"/>.
+        /// </summary>
+        /// <returns>the corresponding <see cref="MailboxAddress"/>.</returns>
+        internal MailboxAddress AsMailboxAddress() => new MailboxAddress(DisplayName, Address);
```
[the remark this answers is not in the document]
bjss-chris: I remember pondering this decision for exactly the reason you've stated but now I can't remember why I made the decision I did!

I'm going to refactor those methods to extension methods – that way the `EmailAddress` and `EmailMessage` classes will no longer have any knowledge of their MimeKit equivalents, we get to keep the helpful syntax and all the functionality is contained in a single class that is clearly specific to the MailKit/MimeKit implementation.

If we're still not sure about the extension method version then I'm not opposed to refactoring again.

src/NHSD.BuyingCatalogue.Organisations.Api/Services/EmailAddress.cs
bjss-chris: Json.Net does support deserialization using a non-default constructor but the configuration system uses `System.Text.Json` as far as I'm aware (and I don't think it's configurable). I believe it is possible to deserialize an immutable instance using `System.Text.Json` by writing a custom converter.

I considered having a separate class for the configuration but I felt it was going to result in duplication and breaking the DRY principle.

I'm not sure it's worth spending much more time on this now but it's certainly something we could come back to later – it would be good just to get the basic functionality in for now.

Suggestions are always welcome though 👍

PR #90, opened 2020-03-26
5655: Set the username in create user POST endpoint

assoomro:


tests/NHSD.BuyingCatalogue.Identity.Api.IntegrationTests/Utils/Database.cs
```diff
@@ -17,9 +17,9 @@ public async Task Reset(IConfigurationRoot config)
             using IDbConnection databaseConnection = new SqlConnection(config.GetConnectionString("CatalogueUsersAdmin"));
             await databaseConnection.ExecuteAsync("ALTER ROLE db_datareader ADD MEMBER NHSD;");
             await databaseConnection.ExecuteAsync("ALTER ROLE db_datawriter ADD MEMBER NHSD;");
-            await databaseConnection.ExecuteAsync("DELETE FROM Organisations");
-            await databaseConnection.ExecuteAsync("DELETE FROM AspNetUsers WHERE SecurityStamp = 'TestUser'");
-            await databaseConnection.ExecuteAsync("DELETE FROM AspNetUsers WHERE UserName IS NULL");
+            await databaseConnection.ExecuteAsync("DELETE FROM Organisations;");
+            await databaseConnection.ExecuteAsync("DELETE FROM AspNetUsers WHERE SecurityStamp = 'TestUser';");
```
sam-roberts-bjss: Really think we need to review this one. It is strange to have two delete statements for the same table. We might as well delete everything as the tests are not setup to run in parallel yet?

assoomro: My understanding of this is as follows:
- 3 accounts are injected on start up - Alice (non authority), Bob (authority) and an Agency user (based on env var)
- When we run our tests, the tests insert users as we want (thanks @CallumBradburyBJSS ) - those users have the SecurityStamp set to 'TestUser' - therefore we delete them before every scenario to start with a clean slate.
- The [tests I've written earlier](https://github.com/nhs-digital-gp-it-futures/BuyingCatalogueIdentity/commit/3c4b57d056ca207f24b05e215f2c082d872695fa) call the Create user endpoint, which is equivalent to how we would _actually add users_ and by the end of my tests, the're will be some tests left in the database that don't have the SecurityStamp set to 'TestUser' therefore won't be deleted. The statement I've added (the 2nd statement you're referring to) is there to remove those, but keep Bob, Alice and agency user in.

The solution to this would in my opinion be - refactor existing tests to stop using Bob/Alice/Agency user (unless needed somewhere), and delete the whole table after each scenario. 

Should we raise a tech debt ticket for this then @sam-roberts-bjss ?

sam-roberts-bjss: Yeah good shout. There are a few options we could try but that is a good first step. We want to be careful of making tech debt the dumping ground. It is already quite big. In this instance the task at hand is related more to the creating of a user, so yeah lets raise a task.

PR #93, opened 2020-03-27
5695: Fix docker build issue - missing Identity.Common

sam-roberts-bjss:


src/NHSD.BuyingCatalogue.Organisations.Api/Dockerfile
```diff
@@ -1,14 +1,15 @@
-#See https://aka.ms/containerfastmode to understand how Visual Studio uses this Dockerfile to build your images for faster debugging.
-
 FROM mcr.microsoft.com/dotnet/core/aspnet:3.1-buster-slim AS base
+RUN apt update && apt install iproute2 -y 
 WORKDIR /app
 EXPOSE 80
-EXPOSE 443
```
assoomro: I'm not 100% sure on this one, but I think organisations might need this. I think @stuart-bass-bjss would be able to confirm, but since it has already been in, I wouldn't touch it as might might mess with the deployments.

PR #94, opened 2020-03-27
5722: Refactored integration tests

assoomro:
to stop relying on Bob & Alice accounts being there.

### DIFF
--- a/tests/NHSD.BuyingCatalogue.Identity.Api.IntegrationTests/Utils/Database.cs
+++ b/tests/NHSD.BuyingCatalogue.Identity.Api.IntegrationTests/Utils/Database.cs
@@ -1,5 +1,4 @@
-﻿using System.Collections.Generic;
-using System.Data;
+﻿using System.Data;
 using System.Data.SqlClient;
 using System.Threading.Tasks;
 using Dapper;
@@ -18,8 +17,7 @@ namespace NHSD.BuyingCatalogue.Identity.Api.IntegrationTests.Utils
             await databaseConnection.ExecuteAsync("ALTER ROLE db_datareader ADD MEMBER NHSD;");
             await databaseConnection.ExecuteAsync("ALTER ROLE db_datawriter ADD MEMBER NHSD;");
             await databaseConnection.ExecuteAsync("DELETE FROM Organisations;");
-            await databaseConnection.ExecuteAsync("DELETE FROM AspNetUsers WHERE SecurityStamp = 'TestUser';");
-            await databaseConnection.ExecuteAsync("DELETE FROM AspNetUsers WHERE LockoutEnabled != 1;");
+            await databaseConnection.ExecuteAsync("DELETE FROM AspNetUsers;");
         }
 
         public static async Task RemoveReadRole(string connectionString)
--- a/tests/NHSD.BuyingCatalogue.Identity.Api.IntegrationTests/Utils/Database.cs
+++ b/tests/NHSD.BuyingCatalogue.Identity.Api.IntegrationTests/Utils/Database.cs
@@ -17,7 +17,7 @@ namespace NHSD.BuyingCatalogue.Identity.Api.IntegrationTests.Utils
             await databaseConnection.ExecuteAsync("ALTER ROLE db_datareader ADD MEMBER NHSD;");
             await databaseConnection.ExecuteAsync("ALTER ROLE db_datawriter ADD MEMBER NHSD;");
             await databaseConnection.ExecuteAsync("DELETE FROM Organisations;");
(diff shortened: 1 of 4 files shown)


tests/NHSD.BuyingCatalogue.Identity.Api.IntegrationTests/Utils/Database.cs
```diff
@@ -18,8 +17,7 @@ public async Task Reset(IConfigurationRoot config)
             await databaseConnection.ExecuteAsync("ALTER ROLE db_datareader ADD MEMBER NHSD;");
             await databaseConnection.ExecuteAsync("ALTER ROLE db_datawriter ADD MEMBER NHSD;");
             await databaseConnection.ExecuteAsync("DELETE FROM Organisations;");
-            await databaseConnection.ExecuteAsync("DELETE FROM AspNetUsers WHERE SecurityStamp = 'TestUser';");
-            await databaseConnection.ExecuteAsync("DELETE FROM AspNetUsers WHERE LockoutEnabled != 1;");
+            await databaseConnection.ExecuteAsync("DELETE FROM AspNetUsers;");
```
CallumBradburyBJSS: should we really be deleting all users? I liked us just deleting the ones we'd added, seems excessive to do otherwise

assoomro: I liked the 'newly created' selection too. However after [this PR](https://github.com/nhs-digital-gp-it-futures/BuyingCatalogueIdentity/commit/3c4b57d056ca207f24b05e215f2c082d872695fa) our integration tests started adding 'real' users which don't have  "TestUser" as their Security stamp (its a guid) and thus,  the 'LockoutEnabled' query was born.

The LockoutEnabled was a temporary fix in my eyes as its not clear which users does that affect (Bob, Alice and EA User have it set to true, so they don't get deleted).

After weighing
A) Keep 'LockoutEnabled' query which keeps users, but is vague
&
B) Delete all users and lose the option to log in after running integration tests, but be clear

I went with B as clarity > convenience, assuming most people wouldn't run the integration tests and then try to log in using alice/bob/ea user.

I would, however, like to know your opinion on what the best approach would be and whether it is justifiable to keep the 'LockoutEnabled' query or not.

assoomro: As per our discussion:

There are few more options at hand:

1. Keep a track of added users through scenario context and delete them with a clean-up step
2. At the beginning of our tests, query the db for its users at that time and at the end of the tests, delete all newly added.
3. Change Bob&Alice's Security stamp in the sql script & don't delete them based on this stamp
4. At the end of the test run, delete all users except for Bob, Alice and EA User by some identifier (probably email)

Solution #1 & #2 would be problematic if devs exit the tests prematurely as this would keep the newly added users in the db, causing problems in the subsequent runs.
Solution #3 seemed promising, however meddling with the EA user doesn't seem appropriate as it would involve introducing a new env variable and the platform team would need to be aware of this and it would increase their workload unnecessarily. 

After a lot of considerations, @CallumBradburyBJSS and I agreed that Solution #4, however inelegant, seems to be the best approach to take at the moment.

tests/NHSD.BuyingCatalogue.Identity.Api.IntegrationTests/Utils/Database.cs
```diff
@@ -18,8 +17,7 @@ public async Task Reset(IConfigurationRoot config)
             await databaseConnection.ExecuteAsync("ALTER ROLE db_datareader ADD MEMBER NHSD;");
             await databaseConnection.ExecuteAsync("ALTER ROLE db_datawriter ADD MEMBER NHSD;");
             await databaseConnection.ExecuteAsync("DELETE FROM Organisations;");
-            await databaseConnection.ExecuteAsync("DELETE FROM AspNetUsers WHERE SecurityStamp = 'TestUser';");
-            await databaseConnection.ExecuteAsync("DELETE FROM AspNetUsers WHERE LockoutEnabled != 1;");
+            await databaseConnection.ExecuteAsync("DELETE FROM AspNetUsers WHERE [Email] NOT IN ('user@agency.com', 'AliceSmith@email.com', 'BobSmith@email.com');");
```
CallumBradburyBJSS: Is this case sensitive?

assoomro: This casing is replicating the one in the DB. It feels like an overkill but I suppose I could refactor it to use NormalisedEmail instead.

Thoughts?

PR #102, opened 2020-04-01
5605: User Details GET

RuthBerry96:


src/NHSD.BuyingCatalogue.Organisations.Api/Controllers/UsersController.cs
```diff
@@ -80,5 +81,28 @@ public async Task<ActionResult> CreateUserAsync(Guid organisationId, CreateUserR
 
             return Ok();
         }
+
+        [Route("users/{userId}")]
```
bjss-chris, line 85: If we're following the convention used for the organizations routes, this probably ought to be `[Route("api/v1/users/{userId}")]`.

bjss-chris: As an aside, I think we probably ought to have a think about our routing and whether all our action methods currently reside in the most appropriate controllers. For example, the current routing implies that two of the methods in this class probably ought to live in the organizations controller.

I don't think we need to do anything now as part of this change – this comment is more of a note as I think it's probably something that might be worthwhile discussing.

src/NHSD.BuyingCatalogue.Organisations.Api/Controllers/UsersController.cs
```diff
@@ -80,5 +81,28 @@ public async Task<ActionResult> CreateUserAsync(Guid organisationId, CreateUserR
...
+        public async Task<ActionResult> GetUserById(string userId)
+        {
+            var user = await _usersRepository.GetUserById(userId);
+
+            if (user is null)
+            {
+                return NotFound();
+            }
+
+            var userViewModel = new GetUserViewModel()
```
bjss-chris, line 96: The brackets can be omitted here (they're not required with object initializers).

src/NHSD.BuyingCatalogue.Organisations.Api/Controllers/UsersController.cs
```diff
@@ -80,5 +81,28 @@ public async Task<ActionResult> CreateUserAsync(Guid organisationId, CreateUserR
 
             return Ok();
         }
+
+        [Route("users/{userId}")]
+        [HttpGet]
+        public async Task<ActionResult> GetUserById(string userId)
```
bjss-chris, line 87: The return result can make use of the generic version of `ActionResult`, i.e. `Task<ActionResult<GetUserViewModel>>`. One of the benefits of that is  it reduces the number of Swagger annotations required (if we get round to enabling it).

src/NHSD.BuyingCatalogue.Organisations.Api/Controllers/UsersController.cs
```diff
@@ -80,5 +81,28 @@ public async Task<ActionResult> CreateUserAsync(Guid organisationId, CreateUserR
...
+            var userViewModel = new GetUserViewModel()
+            {
+                Name = $"{user.FirstName} {user.LastName}",
+                PhoneNumber = user.PhoneNumber,
+                EmailAddress =  user.Email,
+                Disabled = user.Disabled,
+                PrimaryOrganisationId = user.PrimaryOrganisationId
+            };
+
+            return Ok(userViewModel);
```
bjss-chris, line 105: We're not actually returning a view here so I think the variable (and corresponding) type would be better without the `ViewModel` suffix.

src/NHSD.BuyingCatalogue.Organisations.Api/Repositories/UsersRepository.cs
```diff
@@ -32,5 +32,10 @@ public async Task CreateUserAsync(ApplicationUser user)
             _context.Users.Add(user);
             await _context.SaveChangesAsync();
         }
+
+        public async Task<ApplicationUser> GetUserById(string userId)
+        {
+            return await _context.Users.FirstOrDefaultAsync(x => x.Id == userId);
+        }
```
bjss-chris, line 39: As a result of some of the higher-level design choices we're starting to see more duplication. For example, this performs the same function as `UserRepository.FindByIdAsync` in the Identity project.

This implementation is absolutely fine – this is really just another prompt that I think we're getting to the point we might want to discuss the impact of some of our design choices so far and potential options.

src/NHSD.BuyingCatalogue.Organisations.Api/ViewModels/Users/GetUserViewModel.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System;
+
+namespace NHSD.BuyingCatalogue.Organisations.Api.ViewModels.Users
+{
+    public sealed class GetUserViewModel
```
bjss-chris, line 5: As above, as we're not using this model in the context of a view I'm not sure it should have the `ViewModel` suffix (as it's potentially misleading).

tests/NHSD.BuyingCatalogue.Identity.Api.IntegrationTests/Features/Get User.5605.feature
```diff
@@ -0,0 +1,54 @@
...
+	Then a response with status code 401 is returned
+
+@5605
+Scenario: 5. Service Failure
+	Given an user is logged in
+		| Username          | Password | Scope        |
+		| authority@doe.com | yolo     | Organisation |
+	Given the call to the database will fail
+	When a GET request is made for a user with id 234
+	Then a response with status code 500 is returned
```
bjss-chris, line 54: File should end with a newline to keep git happy.

tests/NHSD.BuyingCatalogue.Identity.Api.IntegrationTests/Steps/UserSteps.cs
```diff
@@ -103,6 +103,37 @@ public async Task WhenAUserIsCreated(string organisationName, Table table)
             _response.Result = await client.PostAsync(new Uri($"{_organisationUrl}/{organisationId}/users"), content);
         }
 
+        [When(@"a GET request is made for a user with id (.*)")]
+        public async Task WhenAGETRequestIsMadeForAUserWithId(string userId)
+        {
+            using var client = new HttpClient();
+            client.SetBearerToken(_context.Get(ScenarioContextKeys.AccessToken, ""));
```
bjss-chris, line 110: Personally I prefer `string.Empty` because it leaves me in no doubt about intent but `""` is used elsewhere in this class so feel free to leave as is.

(There are currently no code standards around this either as far as I'm aware but I suspect it would be another developer's prerogative one anyway.)

tests/NHSD.BuyingCatalogue.Identity.Api.IntegrationTests/Steps/UserSteps.cs
```diff
@@ -103,6 +103,37 @@ public async Task WhenAUserIsCreated(string organisationName, Table table)
...
+        [Then(@"a user is returned with the following values")]
+        public async Task ThenAUserIsReturnedWithTheFollowingValues(Table table)
+        {
+            var expected = table.CreateSet<ExpectedGetUserTable>().First();
+
+            var organisationId = GetOrganisationIdFromName(expected.OrganisationName);
+
+            expected.PrimaryOrganisationId = organisationId;
+
+            var response = (await _response.ReadBody());
```
bjss-chris, line 123: The additional brackets aren't required here.

tests/NHSD.BuyingCatalogue.Organisations.Api.UnitTests/Controllers/UsersControllerTests.cs
```diff
@@ -171,5 +171,51 @@ private static (ApplicationUser RepoUser, OrganisationUserViewModel ExpectedUser
                 }
             );
         }
+
+        [Test]
+        public async Task GetUserById_WithExistingUserId_ReturnsOkayObjectResult()
```
bjss-chris, line 176: We're testing more than just the result type here so probably best tweaking the name.

tests/NHSD.BuyingCatalogue.Organisations.Api.UnitTests/TestContexts/UsersControllerTestContext.cs
```diff
@@ -32,6 +34,8 @@ private UsersControllerTestContext()
 
         internal Mock<IRegistrationService> RegistrationServiceMock { get; set; }
 
+        public ApplicationUser GetUserById { get; set; }
```
bjss-chris, line 37: I think this might be better just named `User` as the current name is slightly confusing at first glance (and it has the potential for later reuse).

PR #103, opened 2020-04-01
5697: Add db dependency to healthchecks

assoomro:


src/NHSD.BuyingCatalogue.Organisations.Api/Startup.cs
```diff
@@ -37,6 +37,7 @@ public Startup(IConfiguration configuration)
         // This method gets called by the runtime. Use this method to add services to the container.
         public void ConfigureServices(IServiceCollection services)
         {
+        	var connectionString = Configuration.GetConnectionString("CatalogueUsers");
```
bjss-chris, line 40: Alignment/spacing – I suspect there's a tab here rather than spaces.

tests/NHSD.BuyingCatalogue.Identity.Api.IntegrationTests/Steps/HealthCheckSteps.cs
```diff
@@ -22,8 +22,8 @@ public HealthChecksSteps(Response response, ScenarioContext context, Settings se
             _context["organisationBaseUrl"] = _settings.OrganisationApiBaseUrl;
         }
 
-        [Given(@"The Smtp Server is (up|down)")]
-        public void GivenTheSmptSeverIsInState(string state)
+        [Given(@"The (Smtp|Database) Server is (up|down)")]
+        public void GivenTheServerIsInState(string server, string state)
```
bjss-chris, line 26: The `server` parameter isn't currently used – should it be?

tests/NHSD.BuyingCatalogue.Identity.Api.IntegrationTests/Features/Organisations Healthcheck.3538.feature
```diff
@@ -14,3 +14,15 @@ Scenario: 2. Smtp Server is down
 	Given The Smtp Server is down
 	When the dependency health-check endpoint is hit
 	Then the response will be Unhealthy
+
+@5648
+Scenario: 3. Database Server is up
```
sam-roberts-bjss, line 19: Please can you confirm what the test is doing here? It looks to me like we change the address of the health check endpoint indicating that the database is down. Would a better approach be to revoke the read and write privileges of the database then access the normal organisation health check URL?

src/NHSD.BuyingCatalogue.Organisations.Api/Startup.cs
sam-roberts-bjss, line 40: Indentation?

tests/NHSD.BuyingCatalogue.Identity.Api.IntegrationTests/Steps/HealthCheckSteps.cs
assoomro: The param is there only to satisfy specflow. The step would be functional without it - being just 'The Server is up/down', however I thought adding 'Smtp/Database' would give a bit more context.

There's a few ways to go about this, however I'm not sure on which one is the best.
1) Have 2 separate steps - The Smtp Server is up/down & The Database Server is up/down. These two steps would do exactly the same thing as the current method.
2) Have 1 step 'The Server is up/down' and would do what the current method does.
3) The current way with a redundant parameter.

1) Feels like violating DRY
2) Feels bit too vague
3) is not the best but better than 1 & 2

What do you think?

sam-roberts-bjss:
Is there a way to check for degradation? When the SMTP is down do we class the Organisations.Api application as down or do we mark it as degraded?

tests/NHSD.BuyingCatalogue.Identity.Api.IntegrationTests/Features/Organisations Healthcheck.3538.feature
assoomro: As per our discussion over the phone - 
The other address points to a 'carbon copy' of the API, with the only difference in them being access to the network - meaning the 'Broken API' can't talk to the db / smtp server.

Revoking r/w privileges wouldn't make the health check fail as the user would still be able to query 'SELECT 1;' and get a healthy response from the db. A way around that would be to drop the user before the test and then restore it afterwards, the same way we do in BAPI, however this approach tends to be more messy.
Given that the simplest way to 'break' the SMTP server is to use the broken API's endpoint, I thought doing the same for the database check would be appropriate as its the simplest solution that works.

src/NHSD.BuyingCatalogue.Organisations.Api/Startup.cs
assoomro: 👍

assoomro: 👍

tests/NHSD.BuyingCatalogue.Identity.Api.IntegrationTests/Steps/HealthCheckSteps.cs
bjss-chris: The given template is just a regex isn't it so can you just remove the capture group, so the attribute becomes `[Given(@"The Smtp|Database Server is (up|down)")]`?

assoomro:
> Is there a way to check for degradation? When the SMTP is down do we class the Organisations.Api application as down or do we mark it as degraded?

That is a great question. I was torn on this one, I think we should decide on this as a team, I will, however leave my thoughts in here for a reference to my original choices.

The organisations API depends mainly on the db - without the db, its not functional, therefore it makes sense to return Unhealthy if the db is down.
As for the SMTP server - the API can function without it, only 1 feature depends on it, therefore it should return Degraded when SMTP server is down. 
However, on second thought - if the SMTP server is down, we are unable to create buyer accounts because the 'create account' journey can never be completed. 
When I thought about it that way, it made more sense to 'bump up' the seriousness of the dependency and have it return Unhealthy if there's something wrong with the SMTP server.

But as I said, this should be decided by the team, not me, so please leave your thoughts below.

tests/NHSD.BuyingCatalogue.Identity.Api.IntegrationTests/Steps/HealthCheckSteps.cs
assoomro: That's fantastic! I didn't know about that feature, thank you Chris, I'll change it to that.

bjss-chris:
> Is there a way to check for degradation? When the SMTP is down do we class the Organisations.Api application as down or do we mark it as degraded?
> 
> However, on second thought - if the SMTP server is down, we are unable to create buyer accounts because the 'create account' journey can never be completed.
> When I thought about it that way, it made more sense to 'bump up' the seriousness of the dependency and have it return Unhealthy if there's something wrong with the SMTP server.

My initial thought is probably that the service is degraded when the SMTP server is down. While this is going to impact the user creation story other services the API provides will still be available, such as viewing organizations and their users. However, this also ties in to the outstanding requirements around SMTP/e-mail failures...

src/NHSD.BuyingCatalogue.Organisations.Api/Startup.cs
```diff
@@ -82,6 +83,13 @@ public void ConfigureServices(IServiceCollection services)
                     "self",
                     () => HealthCheckResult.Healthy(),
                     new[] { HealthCheckTags.Live })
+                .AddSqlServer(
```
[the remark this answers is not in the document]
assoomro: I quite like the idea of tidying it up with extension methods. I'll do that.

sam-roberts-bjss:
> > Is there a way to check for degradation? When the SMTP is down do we class the Organisations.Api application as down or do we mark it as degraded?
> > However, on second thought - if the SMTP server is down, we are unable to create buyer accounts because the 'create account' journey can never be completed.
> > When I thought about it that way, it made more sense to 'bump up' the seriousness of the dependency and have it return Unhealthy if there's something wrong with the SMTP server.
> 
> My initial thought is probably that the service is degraded when the SMTP server is down. While this is going to impact the user creation story other services the API provides will still be available, such as viewing organizations and their users. However, this also ties in to the outstanding requirements around SMTP/e-mail failures...

I agree only one feature of the application will be effected so seems sensible to me to mark it as degraded. Unfortunately this will effect how we've implemented the health check integration tests. I am not sure with the current implementation of using the "Broken" organisations API container, how it will work.

assoomro:
> > > Is there a way to check for degradation? When the SMTP is down do we class the Organisations.Api application as down or do we mark it as degraded?
> > > However, on second thought - if the SMTP server is down, we are unable to create buyer accounts because the 'create account' journey can never be completed.
> > > When I thought about it that way, it made more sense to 'bump up' the seriousness of the dependency and have it return Unhealthy if there's something wrong with the SMTP server.
> > 
> > 
> > My initial thought is probably that the service is degraded when the SMTP server is down. While this is going to impact the user creation story other services the API provides will still be available, such as viewing organizations and their users. However, this also ties in to the outstanding requirements around SMTP/e-mail failures...
> 
> I agree only one feature of the application will be effected so seems sensible to me to mark it as degraded. Unfortunately this will effect how we've implemented the health check integration tests. I am not sure with the current implementation of using the "Broken" organisations API container, how it will work.

Seems like Degraded is the way to go forwards. Sam has a point. The current implementation would not allow for an easy transition.

There would be a few options:
1) 1 extra broken service (3 in total, 1 healthy, 2 broken) - for db / smtp separately
2) Using docker CLI from the integration tests to stop / start containers on the go
3) Reloading the config file - I dont know how exactly would this work but @sam-roberts-bjss does.

I think it would boil down to selecting the lesser evil.

We could also leave it as it is for now and add a tech-debt ticket to address those problems at a later date.

Opinions?

src/NHSD.BuyingCatalogue.Organisations.Api/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,41 @@
...
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Diagnostics.HealthChecks;
+using NHSD.BuyingCatalogue.Identity.Common.Constants;
+using NHSD.BuyingCatalogue.Organisations.Api.Settings;
+
+namespace NHSD.BuyingCatalogue.Organisations.Api.Extensions
+{
+    internal static class ServiceCollectionExtensions
+    {
+        public static IServiceCollection RegisterHealthChecks(this IServiceCollection services, string connectionString, SmtpSettings smtpSettings)
```
sam-roberts-bjss, line 12: Can `connectionString` or `smtpSettings` parameters be null? I think this one will come up in the coding standards meeting tomorrow but we made @bjss-chris change his code to guard & fail fast with parameters that should not be null.

src/NHSD.BuyingCatalogue.Organisations.Api/Startup.cs
```diff
@@ -14,6 +14,7 @@
 using Microsoft.Extensions.Logging;
 using NHSD.BuyingCatalogue.Identity.Common.Constants;
 using NHSD.BuyingCatalogue.Organisations.Api.Data;
+using NHSD.BuyingCatalogue.Organisations.Api.Extensions;
```
bjss-chris, line 17: I think there may be some `using` statements that we no longer require now.

sam-roberts-bjss:
In response to your question above I think option 1 at this stage will be the simplest.

tests/NHSD.BuyingCatalogue.Identity.Api.IntegrationTests/Steps/HealthCheckSteps.cs
bjss-chris: My mistake with the suggestion here. I should have suggested a non-capturing group:

```c#
[Given(@"The (?:Smtp|Database) Server is (up|down)")]
```

src/NHSD.BuyingCatalogue.Organisations.Api/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,45 @@
...
+        {
+            if (connectionString is null) throw new ArgumentNullException(nameof(connectionString));
+
+            if (smtpSettings is null) throw new ArgumentNullException(nameof(smtpSettings));
```
bjss-chris, line 16: Can the `throw` statements be moved to a new line? E.g.

```c#
if (connectionString is null)
    throw new ArgumentNullException(nameof(connectionString));
```
I think this is improves the readability.

assoomro: 👍 I don't have a problem moving them to a new line.

PR #108, opened 2020-04-02
5696: Isapi db healthcheck

assoomro:


src/NHSD.BuyingCatalogue.Identity.Api/Startup.cs
```diff
@@ -131,6 +136,15 @@ public void ConfigureApp(IApplicationBuilder app)
             {
                 endpoints.MapDefaultControllerRoute();
                 endpoints.MapControllers();
+                endpoints.MapHealthChecks("/health/live", new HealthCheckOptions
+                {
+                    Predicate = (healthCheckRegistration) => healthCheckRegistration.Tags.Contains(HealthCheckTags.Live)
```
bjss-chris, line 141: The brackets around the lambda parameter can be omitted here.

src/NHSD.BuyingCatalogue.Identity.Api/Startup.cs
```diff
@@ -131,6 +136,15 @@ public void ConfigureApp(IApplicationBuilder app)
...
+                });
+
+                endpoints.MapHealthChecks("/health/ready", new HealthCheckOptions
+                {
+                    Predicate = (healthCheckRegistration) => healthCheckRegistration.Tags.Contains(HealthCheckTags.Ready)
```
bjss-chris, line 146: The brackets around the lambda parameter can be omitted here.

src/NHSD.BuyingCatalogue.Identity.Api/Startup.cs
assoomro: 🤦‍♂ D'oh! I'll get rid of them.

src/NHSD.BuyingCatalogue.Identity.Api/Startup.cs
assoomro: 🤦‍♂ D'oh! I'll get rid of them

PR #111, opened 2020-04-03
Move e-mail functionality to the common project

bjss-chris:
AB#5615

### DIFF
--- a/src/NHSD.BuyingCatalogue.Identity.Common/Email/EmailAddress.cs
+++ b/src/NHSD.BuyingCatalogue.Identity.Common/Email/EmailAddress.cs
@@ -1,13 +1,13 @@
 ﻿using System;
 
-namespace NHSD.BuyingCatalogue.Organisations.Api.Services
+namespace NHSD.BuyingCatalogue.Identity.Common.Email
 {
     /// <summary>
     /// An e-mail address.
     /// </summary>
-    internal sealed class EmailAddress
+    public sealed class EmailAddress
     {
-        private string _address;
+        private string? _address;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="EmailAddress"/> class.
@@ -23,7 +23,7 @@ namespace NHSD.BuyingCatalogue.Organisations.Api.Services
         /// </summary>
         /// <param name="displayName">An optional display name.</param>
         /// <param name="address"></param>
-        internal EmailAddress(string displayName, string address)
+        public EmailAddress(string displayName, string address)
         {
             DisplayName = displayName;
             Address = address;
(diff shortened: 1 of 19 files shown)


src/NHSD.BuyingCatalogue.Organisations.Api/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -14,7 +14,7 @@ public static IServiceCollection RegisterHealthChecks(this IServiceCollection se
             if (connectionString is null)
                 throw new ArgumentNullException(nameof(connectionString));
 
-            if (smtpSettings is null) 
+            if (smtpSettings is null)
```
CallumBradburyBJSS, line 17: Would adding the NotNull attribute to the parameter work here? I've not messed with it much, but I think it's purpose is to get rid of this kind of thing

src/NHSD.BuyingCatalogue.Identity.Common/Email/EmailAddress.cs
```diff
@@ -1,13 +1,13 @@
...
+namespace NHSD.BuyingCatalogue.Identity.Common.Email
 {
     /// <summary>
     /// An e-mail address.
     /// </summary>
-    internal sealed class EmailAddress
+    public sealed class EmailAddress
     {
-        private string _address;
+        private string? _address;
```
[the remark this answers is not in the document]
bjss-chris: They are except when nullable reference types are enabled as you spotted below :D

src/NHSD.BuyingCatalogue.Organisations.Api/Extensions/ServiceCollectionExtensions.cs
bjss-chris: The `[NotNull]` attribute was introduced in 3.0 as part of the `System.Diagnostics.CodeAnalysis` namespace but I'm not yet sure of its correct/intended usage. It doesn't appear to be that well documented but it appears to be more about outputs than inputs. For example, from the [documentation](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.codeanalysis.notnullattribute?view=netcore-3.1):

> Specifies that an input argument was not null when the call returns.

I don't believe it will do anything to prevent a null parameter being passed in, although it's probably worth investigating the correct/intended usage.

The only other `[NotNull]` attribute I'm aware of is the R# annotation but that obviously only works with ReSharper so it may not be something we want to rely on.

bjss-chris:
Although I've enabled nullable reference types in the common project I didn't want to spend time refactoring the e-mail functionality to take full advantage of that feature at this point – I just wanted to get the core functionality moved. We can refactor to take full advantage of the nullable reference types later.

PR #115, opened 2020-04-06
6072: Add dummy Buyer Users to data setup

assoomro:
If the 'INSERT_TEST_DATA' env var is set to 'True', this adds 30 buyer users belonging to 10 organisations to our Users table.


src/NHSD.BuyingCatalogue.Identity.UserDatabase/Deployment/CreateTestBuyers.sql
```diff
@@ -0,0 +1,40 @@
...
+    (N'd94109ef-6051-4937-8a9e-8599a2707096', N'Larry.Ebert94@gmail.com', N'LARRY.EBERT94@GMAIL.COM', N'Larry.Ebert94@gmail.com', N'LARRY.EBERT94@GMAIL.COM', 0, NULL, N'49bdba08-357a-4204-b3a1-25c73f582ca6', N'146602f2-3090-4543-8edc-3d078dd470e2', N'95740022454', 0, 0, NULL, 0, 0, N'b1e85a4d-364c-4adc-a981-eb21b58a2cc0', N'Buyer', 0, 0, N'Larry', N'Ebert'),
+    (N'db70b19d-5387-4d36-8957-a64dd9d87e66', N'Grover29@yahoo.com', N'GROVER29@YAHOO.COM', N'Grover29@yahoo.com', N'GROVER29@YAHOO.COM', 0, NULL, N'c2c9ca3f-597a-44a8-bc47-75433cf89cdb', N'606403ae-ac27-4a57-bb6f-bd73605357e2', N'74529061674', 0, 0, NULL, 0, 0, N'145cdd34-0bb9-4122-a0bf-1b6fcb0219bb', N'Buyer', 0, 0, N'Grover', N'Nader'),
+    (N'e8fd68d0-5ec0-40d3-a2b1-58cd4769b099', N'Santos33@gmail.com', N'SANTOS33@GMAIL.COM', N'Santos33@gmail.com', N'SANTOS33@GMAIL.COM', 0, NULL, N'c8a655f9-4a64-48d3-b0ea-1b3e07604435', N'fb0ec71c-e961-44a5-9b7e-a8a4beb62b85', N'41140608953', 0, 0, NULL, 0, 0, N'd795a598-26f0-46fb-8a6e-f22cd912d7dc', N'Buyer', 0, 0, N'Santos', N'Kuhlman')
```
bjss-chris, line 39: I'm wondering if we probably ought to use reserved DNS names for our test users to minimize the risk of actually sending an e-mail to a genuine address. These are domains that a real SMTP server should not resolve – see [RFC 2606](https://tools.ietf.org/html/rfc2606#page-2) for more details.

The reserved names are:

- .test
- .example
- .invalid
- .localhost
- example.com
- example.net
- example.org

assoomro:
As per discussion with @sam-roberts-bjss, @talawson05 we don't actually need this. Closing the PR.

PR #118, opened 2020-04-06
Generate password reset token

bjss-chris:
AB#5615


src/NHSD.BuyingCatalogue.Identity.Api/Services/PasswordService.cs
```diff
@@ -0,0 +1,93 @@
...
+        }
+
+        /// <summary>
+        /// Generates a password reset token for the user with the
+        /// provided <paramref name="emailAddress"/>.
+        /// </summary>
+        /// <param name="emailAddress">The e-mail of the user to generate the password
+        /// reset token for.</param>
+        /// <returns>A <see cref="PasswordResetToken"/> if the user was found;
+        /// otherwise, <see lagref="null"/>.</returns>
```
CallumBradburyBJSS, line 46: lagref?

src/NHSD.BuyingCatalogue.Identity.Api/Services/PasswordService.cs
```diff
@@ -0,0 +1,93 @@
...
+        public async Task SendResetEmailAsync(ApplicationUser user, string callback)
+        {
+            if (user is null)
+                throw new ArgumentNullException(nameof(user));
+
+            if (callback is null)
+                throw new ArgumentNullException(nameof(callback));
+
+            if (string.IsNullOrWhiteSpace(callback))
+                throw new ArgumentException($"{callback} must be provided", nameof(callback));
```
CallumBradburyBJSS, line 83: this is going to say " should be provided" as callback will be null when it creates that string.

tests/NHSD.BuyingCatalogue.Identity.Api.UnitTests/Controllers/AccountControllerTests.cs
```diff
@@ -261,5 +263,112 @@ public void Logout_WhenInvalidLogoutId_ShouldThrowArgumentNullException(string l
...
+        {
+            using var controller = new AccountControllerBuilder()
+                .Build();
+
+            controller.ModelState.AddModelError(string.Empty, "Fake error!");
+
+            var expectedViewModel = new ForgotPasswordViewModel();
+            var result = await controller.ForgotPassword(expectedViewModel) as ViewResult;
+
+            Assert.NotNull(result);
```
CallumBradburyBJSS, line 290: Use the FluentAssertion!

result.Should().NotBeNull()

src/NHSD.BuyingCatalogue.Identity.Api/Models/ApplicationUser.cs
```diff
@@ -9,6 +9,8 @@ public sealed class ApplicationUser : IdentityUser
 
         public string LastName { get; set; }
 
+        public string DisplayName => $"{FirstName} {LastName}";
```
CallumBradburyBJSS, line 12: I'd add a .Trim() to this, otherwise if there's no first name it'll be " {LastName}", and if there's no last name it'll be "{FirstName} "

tests/NHSD.BuyingCatalogue.Identity.Api.UnitTests/Services/PasswordServiceTests.cs
```diff
@@ -0,0 +1,240 @@
...
+            await service.SendResetEmailAsync(expectedUser, expectedCallback);
+
+            emailCount.Should().Be(1);
+
+            actualEmailMessage.Should().NotBeNull();
+            actualEmailMessage.Sender.Should().Be(expectedSender);
+            actualEmailMessage.Recipient.Should().BeEquivalentTo(
+                new EmailAddress(expectedUser.DisplayName, expectedUser.Email));
+
+            actualEmailMessage.Subject.Should().Be(expectedSubject);
```
CallumBradburyBJSS, line 235: Hmm.

I was looking at this, and remembering the good old days. I quite like this way of verifying calls, but there's the mockEmailService.Verify() which is supposed to be used for this purpose. I had a go at writing one though, and it looked way more ugly, so maybe this is cool?

src/NHSD.BuyingCatalogue.Identity.Api/Services/PasswordService.cs
bjss-chris: Should be `langref` – I'll fix.

src/NHSD.BuyingCatalogue.Identity.Api/Services/PasswordService.cs
bjss-chris: Should be `nameof(callback)` – I'll fix.

tests/NHSD.BuyingCatalogue.Identity.Api.UnitTests/Controllers/AccountControllerTests.cs
bjss-chris: This is deliberate to prevent any static analysis tools from flagging `result` as may be null.

bjss-chris: Possibly another one to add to the next code standards meeting – do we care about the false positive 'may be null'? While I'm still not convinced by fluent assertions I know I'm outnumbered there so it may be that we stick with the fluent version regardless.

I'll leave it as is for now though.

src/NHSD.BuyingCatalogue.Identity.Api/Models/ApplicationUser.cs
bjss-chris: Although there is no database-level constraint that first name or last name cannot be empty (`NOT NULL` only) we do require a first name and last name in the application (`ApplicationUserValidator.ValidateFirstNameAsync` and `ApplicationUserValidator.ValidateLastNameAsync`) so a `.Trim()` shouldn't be necessary.

However, I don't mind adding one if we want to do 'belt and braces'.

tests/NHSD.BuyingCatalogue.Identity.Api.UnitTests/Services/PasswordServiceTests.cs
bjss-chris: I agree about `Verify()` being ugly here, which is why I went for the callback-based approach.

The other option might have been to implement a custom `IEqualityComparer<T>` for tests. However, I wasn't convinced that would be worthwhile for one test.

PR #120, opened 2020-04-06
6086: Fix email validation bug

sam-roberts-bjss:


src/NHSD.BuyingCatalogue.Organisations.Api/Validators/ApplicationUserValidator.cs
```diff
@@ -16,6 +16,8 @@ public class ApplicationUserValidator : IApplicationUserValidator
         private const int MaximumLastNameLength = 50;
         private const int MaximumEmailLength = 256;
 
+        private static readonly EmailAddressAttribute _emailAddressAttribute = new EmailAddressAttribute();
```
assoomro, line 19: This field's name violates our naming rules, ReSharper suggests `EmailAddressAttribute` instead.

sam-roberts-bjss: For a static readonly?

assoomro: As per our discussion: 
Fault is on my side, my tooling is set to something we haven't agreed on as a team, hence I'm taking back my comment and approving this PR.

bjss-chris: I've never been convinced by the default R# setting of suggesting private static fields should be `PascalCase`. I would always go with `camelCase`, which in this code base means they should also have the superfluous leading underscore as @sam-roberts-bjss has done here..

PR #124, opened 2020-04-06
Fix password reset callback URL

bjss-chris:
AB#6091


src/NHSD.BuyingCatalogue.Identity.Api/Controllers/AccountController.cs
```diff
@@ -114,11 +115,9 @@ public async Task<IActionResult> ForgotPassword(ForgotPasswordViewModel viewMode
             if (resetToken == null)
                 return RedirectToAction(nameof(ForgotPasswordLinkSent));
 
-            var callback = Url.Action(
+            var callback = this.Action(
```
sam-roberts-bjss, line 118: Surprised at the use of an extension here. I thought you could do something like the following which will include the controller name for you?

`var callback = Url.Action(new UrlActionContext { Action = nameof(ResetPassword), , Values = new {}});`

bjss-chris: Your example results in a relative URL, but it looks like this version will generate an absolute URL

```c#
var callback = Url.Action(
    new UrlActionContext
    {
        Action = nameof(ResetPassword),
        Protocol = Request.Scheme,
        Values = new { resetToken.Token, Email = viewModel.EmailAddress }
    });
```

I'll switch to that as it means we can get rid of the extension method and associated tests.

bjss-chris: Although I'm just going to double-check your suggestion again. An element of doubt has crept in...

PR #130, opened 2020-04-08
6130: GET org by ods code endpoint with canned data

assoomro:


src/NHSD.BuyingCatalogue.Organisations.Api/Controllers/OdsController.cs
```diff
@@ -0,0 +1,44 @@
...
+    public sealed class OdsController : Controller
+    {
+        [HttpGet]
+        [Route("{odsCode}")]
+        public ActionResult GetByOdsCodeAsync(string odsCode)
+        {
+            if (odsCode is null)
+                throw new ArgumentNullException(nameof(odsCode));
+
+            if (string.IsNullOrWhiteSpace(odsCode))
```
sam-roberts-bjss, line 20: There is a case here for rather than returning an exception, the code could return a 404 notfound status instead. The code will eventually communicate with an external system, there maybe times when an empty or whitespace is a valid value. Unlikely in this instance but it will be up to the "repository" to decide that. This I believe would also be consistent with our other GET endpoints.

src/NHSD.BuyingCatalogue.Organisations.Api/Controllers/OdsController.cs
```diff
@@ -0,0 +1,44 @@
...
+                throw new ArgumentException($"{nameof(odsCode)} cannot be empty.", nameof(odsCode));
+
+            // Canned data
+            return Ok(new OrganisationViewModel
```
sam-roberts-bjss, line 24: Although the data we are returning is the same as the organisation view model I consider this endpoint as a separate query. Further down the line we might want to include additional information from the ODS endpoint but not effect the organisations endpoints, so I think it would be better to use a different "viewmodel" contract e.g OdsViewModel. This keeps the concerns separate.

src/NHSD.BuyingCatalogue.Organisations.Api/Controllers/OdsController.cs
```diff
@@ -0,0 +1,44 @@
+﻿using System;
+using Microsoft.AspNetCore.Mvc;
+using NHSD.BuyingCatalogue.Organisations.Api.ViewModels.Organisations;
+
+namespace NHSD.BuyingCatalogue.Organisations.Api.Controllers
+{
+    //[Authorize(Policy = Policy.CanAccessOrganisation)]
```
sam-roberts-bjss, line 7: I was expecting this `[Authorize(Policy = Policy.CanAccessOrganisation)]` attribute to be included. This will ensure a user with the right privileges can access this data.

assoomro: My reasoning behind leaving it without auth was to make it easier for the FE folk to access this while the data is still canned.

If you feel like it should be 'protected' to start with, I'll un-comment line 5 :]

src/NHSD.BuyingCatalogue.Organisations.Api/Controllers/OdsController.cs
assoomro: Good point, I agree.

src/NHSD.BuyingCatalogue.Organisations.Api/Controllers/OdsController.cs
assoomro: Interesting point you're making. 

I agree, a 404 would be more appropriate here.

src/NHSD.BuyingCatalogue.Organisations.Api/Controllers/OdsController.cs
```diff
@@ -0,0 +1,46 @@
...
+            if (odsCode is null)
+                throw new ArgumentNullException(nameof(odsCode));
+
+            if (string.IsNullOrWhiteSpace(odsCode))
+                return NotFound();
+
+            // Canned data
+            return Ok(new OdsViewModel
+            {
+                OrganisationId = Guid.NewGuid(),
```
sam-roberts-bjss, line 28: I don't think we need the organisationId as for this endpoint we are treating the ODS code as the unique ID.

assoomro: That is right, had a look at the updated JSON contract and we're not returning the org id. I'll remove it from the OdsViewModel. 
Good spot 👍

PR #131, opened 2020-04-08
6133: create org POST endpoint with canned data

assoomro:


src/NHSD.BuyingCatalogue.Organisations.Api/ViewModels/Organisations/CreateOrganisationResponseViewModel.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System.Collections.Generic;
+using NHSD.BuyingCatalogue.Organisations.Api.ViewModels.Messages;
+
+namespace NHSD.BuyingCatalogue.Organisations.Api.ViewModels.Organisations
+{
+    public sealed class CreateOrganisationResponseViewModel
```
CallumBradburyBJSS, line 6: We said we were going to call these Models instead of ViewModels, not sure if that's down as tech debt though - worth considering

assoomro: I remember our discussion on Friday, however I couldn't remember which naming scheme we agreed on, thus I left it as is above.

PR #132, opened 2020-04-08
6204: Add 'ManageOrganisation' policy

assoomro:


assoomro:
Not ready for a merge yet.

src/NHSD.BuyingCatalogue.Organisations.Api/Models/Policy.cs
```diff
@@ -3,6 +3,7 @@
     internal static class Policy
     {
         public const string CanAccessOrganisation = "CanAccessOrganisation";
```
sam-roberts-bjss, line 5: Just to confirm, did we agree these should be plural?

assoomro: We totally did. Gonna fix that :]

PR #139, opened 2020-04-14
6166: Implement GET Ods Organisation by ods Code endpoint

assoomro:


src/NHSD.BuyingCatalogue.Organisations.Api/ViewModels/Organisations/OdsOrganisationViewModel.cs
```diff
@@ -1,6 +1,6 @@
 ﻿namespace NHSD.BuyingCatalogue.Organisations.Api.ViewModels.Organisations
 {
-    public sealed class OdsViewModel
+    public sealed class OdsOrganisationViewModel
```
bjss-chris, line 3: I think we agreed we weren't going to call these `ViewModel` where they aren't used in the context of a view. However, I don't think we agreed what we would call them instead! As I suspect we'll have a tech debt task to rename these I guess this can be left as is for now.

tests/NHSD.BuyingCatalogue.Organisations.Api.UnitTests/Builders/OdsOrganisationBuilder.cs
```diff
@@ -0,0 +1,76 @@
...
+            _isActive = isActive;
+            return this;
+        }
+
+        internal OdsOrganisationBuilder WithIsBuyerOrganisation(bool isBuyerOrganisation)
+        {
+            _isBuyerOrganisation = isBuyerOrganisation;
+            return this;
+        }
+        internal OdsOrganisation Build()
```
bjss-chris, line 63: Just missing white space in between `}` and the method (if we keep the method).

tests/NHSD.BuyingCatalogue.Organisations.Api.UnitTests/Builders/OdsOrganisationBuilder.cs
```diff
@@ -0,0 +1,76 @@
...
+        {
+            _isActive = isActive;
+            return this;
+        }
+
+        internal OdsOrganisationBuilder WithIsBuyerOrganisation(bool isBuyerOrganisation)
+        {
+            _isBuyerOrganisation = isBuyerOrganisation;
+            return this;
+        }
```
bjss-chris, line 62: I'm not sure it's worth including all of these methods if none of them are currently required.

tests/NHSD.BuyingCatalogue.Organisations.Api.UnitTests/Controllers/OdsControllerTests.cs
```diff
@@ -0,0 +1,98 @@
...
+                .Build();
+
+            await controller.GetByOdsCodeAsync(odsCode);
+
+            odsRepositoryMock.Verify(x => x.GetBuyerOrganisationByOdsCode(odsCode), Times.Once);
+        }
+    }
+}
+
+
```
bjss-chris, line 98: Surplus newlines.

tests/NHSD.BuyingCatalogue.Organisations.Api.UnitTests/Controllers/OdsControllerTests.cs
```diff
@@ -0,0 +1,98 @@
...
+using NHSD.BuyingCatalogue.Organisations.Api.Repositories;
+using NHSD.BuyingCatalogue.Organisations.Api.UnitTests.Builders;
+using NUnit.Framework;
+
+namespace NHSD.BuyingCatalogue.Organisations.Api.UnitTests.Controllers
+{
+
+    [TestFixture]
+    [Parallelizable(ParallelScope.All)]
+    public sealed class OdsControllerTests
```
bjss-chris, line 17: (Non-essential) but can be `internal`.

src/NHSD.BuyingCatalogue.Organisations.Api/Controllers/OdsController.cs
```diff
@@ -12,31 +15,46 @@ namespace NHSD.BuyingCatalogue.Organisations.Api.Controllers
...
         [HttpGet]
         [Route("{odsCode}")]
-        public ActionResult GetByOdsCodeAsync(string odsCode)
+        public async Task<ActionResult> GetByOdsCodeAsync(string odsCode)
         {
             if (odsCode is null)
                 throw new ArgumentNullException(nameof(odsCode));
 
             if (string.IsNullOrWhiteSpace(odsCode))
                 return NotFound();
```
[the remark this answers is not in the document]
assoomro: image

As per the image above, ods is fine with 'code' being empty. I have to admit I have no idea how ods handles other whitespace characters, however I think we shouldn't be throwing. 

Why shouldn't we be throwing?
I think because of the user experience. Giving a 500 back when they sent a 'space' character seems worse for the UX than returning a 404.

src/NHSD.BuyingCatalogue.Organisations.Api/Repositories/IOdsRepository.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System.Threading.Tasks;
+using NHSD.BuyingCatalogue.Organisations.Api.Models;
+
+namespace NHSD.BuyingCatalogue.Organisations.Api.Repositories
+{
+    public interface IOdsRepository
+    {
+        Task<OdsOrganisation> GetBuyerOrganisationByOdsCode(string odsCode);
```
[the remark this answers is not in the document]
assoomro: We did indeed, I'll correct that. 👍

src/NHSD.BuyingCatalogue.Organisations.Api/ViewModels/Organisations/OdsOrganisationViewModel.cs
assoomro: 😄 I'm glad you've mentioned this. I'm aware of our meeting and what we agreed on, however when it came to naming it, I didn't want to pave the way forward without consulting with the team! Hence why I left it with the wrong naming scheme. I was counting on the TD task to rename all of them in one fell swoop, as we probably don't want to have half of 'ViewModels' named one way and the rest a different way.

I'll be leaving it as is for now until we rename them 👍

tests/NHSD.BuyingCatalogue.Organisations.Api.UnitTests/Builders/OdsOrganisationBuilder.cs
assoomro: 👍

tests/NHSD.BuyingCatalogue.Organisations.Api.UnitTests/Builders/OdsOrganisationBuilder.cs
assoomro: Good point. I suppose a more minimalistic, sort of write-it-when-you-need-it approach would be more sensible here.

tests/NHSD.BuyingCatalogue.Organisations.Api.UnitTests/Controllers/OdsControllerTests.cs
assoomro: 👍

tests/NHSD.BuyingCatalogue.Organisations.Api.UnitTests/Controllers/OdsControllerTests.cs
assoomro: 👍

src/NHSD.BuyingCatalogue.Organisations.Api/Repositories/OdsRepository.cs
```diff
@@ -0,0 +1,113 @@
...
+namespace NHSD.BuyingCatalogue.Organisations.Api.Repositories
+{
+    internal sealed class OdsRepository : IOdsRepository
+    {
+        private readonly string _buyerOrganisationRoleId;
+        private readonly string _odsApiBaseUrl;
+
+        public OdsRepository(string odsApiBaseUrl)
+        {
+            _buyerOrganisationRoleId = "RO98";
```
bjss-chris, line 18: Did you consider moving the role ID(s) to config?

src/NHSD.BuyingCatalogue.Organisations.Api/Repositories/OdsRepository.cs
```diff
@@ -0,0 +1,113 @@
...
+
+        private static bool IsActive(OdsResponseOrganisation organisation)
+        {
+            return organisation.Status.Equals("Active", StringComparison.OrdinalIgnoreCase);
+        }
+
+        private bool IsBuyerOrganisation(OdsResponseOrganisation organisation)
+        {
+            return organisation.Roles.Role.Any(role =>
+                 role.id.Equals(_buyerOrganisationRoleId, StringComparison.OrdinalIgnoreCase));
```
bjss-chris, line 52: I think this might look better if the entire lambda was moved to the new line, like this:

```c#
return organisation.Roles.Role.Any(
    role => role.id.Equals(_buyerOrganisationRoleId, StringComparison.OrdinalIgnoreCase));
```
However, it's up to you.

src/NHSD.BuyingCatalogue.Organisations.Api/Repositories/OdsRepository.cs
```diff
@@ -0,0 +1,113 @@
...
+        private class OdsResponseAddress
+        {
+            public string AddrLn1 { get; set; }
+            public string AddrLn2 { get; set; }
+            public string AddrLn3 { get; set; }
+            public string AddrLn4 { get; set; }
+            public string Town { get; set; }
+            public string County { get; set; }
+            public string PostCode { get; set; }
+            public string Country { get; set; }
```
bjss-chris, line 100: Is it worth applying consistent formatting here? In your other POCOs you have white space between each property. I'll leave it up to you again.

tests/NHSD.BuyingCatalogue.Organisations.Api.UnitTests/Controllers/OdsControllerTests.cs
```diff
@@ -0,0 +1,96 @@
...
+        [Test]
+        public void GetByOdsCodeAsync_NullOdsCode_ThrowsArgumentNullException()
+        {
+            using var controller = OdsControllerBuilder
+                .Create()
+                .WithGetByOdsCode(null)
+                .Build();
+
+            Assert.ThrowsAsync<ArgumentNullException>(async () => await controller.GetByOdsCodeAsync(null));
+        }
```
bjss-chris, line 76: I'd recommend refactoring this to make use of a local function – there are a few existing examples (such as `Login_LoginViewModel_NullViewModel_ThrowsException` in `AccountControllerTests`).

This avoids the 'access to disposed closure' warning.

tests/NHSD.BuyingCatalogue.Organisations.Api.UnitTests/Controllers/OdsControllerTests.cs
```diff
@@ -0,0 +1,96 @@
...
+
+        [Test]
+        public async Task GetByOdsCodeAsync_VerifyMethodIsCalledOnce_VerifiesMethod()
+        {
+            const string odsCode = "123";
+
+            var odsRepositoryMock = new Mock<IOdsRepository>();
+            odsRepositoryMock.Setup(x => x.GetBuyerOrganisationByOdsCodeAsync(It.IsAny<string>()))
+                .ReturnsAsync(null as OdsOrganisation);
```
bjss-chris, line 85: I think a normal cast might be better here. While `null as OdsOrganisation` works it looks slightly unusual in this context. I think the normal cast looks slightly more familiar here, for example:

```c#
odsRepositoryMock.Setup(x => x.GetBuyerOrganisationByOdsCodeAsync(It.IsAny<string>()))
    .ReturnsAsync((OdsOrganisation)null);
```

tests/NHSD.BuyingCatalogue.Organisations.Api.UnitTests/Repository/OdsRepositoryTests.cs
```diff
@@ -0,0 +1,71 @@
...
+
+            httpTest.RespondWith(status: 200, body: ValidResponseBody);
+
+            var result = await _odsRepository.GetBuyerOrganisationByOdsCodeAsync(OdsCode);
+
+            result.Should().BeOfType<OdsOrganisation>();
+            result.Should().NotBeNull();
+            result.IsActive.Should().BeTrue();
+            result.IsBuyerOrganisation.Should().BeTrue();
+        }
```
bjss-chris, line 35: Do you think it might be worth initializing an expected `OdsOrganisation` and doing a `Should().BeEquivalentTo()` here?

src/NHSD.BuyingCatalogue.Organisations.Api/Repositories/OdsRepository.cs
assoomro: 👍 Now that the amount of Ids has grown, I'll be definitely moving them to config.

src/NHSD.BuyingCatalogue.Organisations.Api/Repositories/OdsRepository.cs
assoomro: Due to change in "the business logic" (buyer organisation is identified as having a primary role id of either RO98 or RO213) I'll be refactoring this anyway 😄

src/NHSD.BuyingCatalogue.Organisations.Api/Repositories/OdsRepository.cs
assoomro: 👍

src/NHSD.BuyingCatalogue.Organisations.Api/Repositories/OdsRepository.cs
```diff
@@ -48,8 +46,12 @@ private static bool IsActive(OdsResponseOrganisation organisation)
...
         private bool IsBuyerOrganisation(OdsResponseOrganisation organisation)
         {
-            return organisation.Roles.Role.Any(role =>
-                 role.id.Equals(_buyerOrganisationRoleId, StringComparison.OrdinalIgnoreCase));
+            return _settings.BuyerOrganisationRoleIds.Contains(GetPrimaryRoleId(organisation));
+        }
+
+        private static string GetPrimaryRoleId(OdsResponseOrganisation organisation)
+        {
+            return organisation.Roles.Role.Where(r => r.primaryRole).Select(r => r.id).FirstOrDefault();
```
bjss-chris, line 54: This is fine. However, are you aware it can also be written as below?

```c#
return organisation.Roles.Role.FirstOrDefault(r => r.primaryRole)?.id;
```
I'd probably go with the version here just because it's slightly simpler. However, it's entirely up to you which version you go with – they both achieve the same result.

tests/NHSD.BuyingCatalogue.Organisations.Api.UnitTests/Controllers/OdsControllerTests.cs
assoomro: 👍

tests/NHSD.BuyingCatalogue.Organisations.Api.UnitTests/Controllers/OdsControllerTests.cs
assoomro: If its easier on the eyes, no problem 😄

src/NHSD.BuyingCatalogue.Organisations.Api/Settings/OdsSettings.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace NHSD.BuyingCatalogue.Organisations.Api.Settings
+{
+    internal sealed class OdsSettings
+    {
+        public string ApiBaseUrl { get; set; }
+        public string[] BuyerOrganisationRoleIds { get; set; }
```
bjss-chris, line 6: I think I'd probably use `HashSet<string>` here. Although it's not going to make any noticeable difference in our case where we currently only have two IDs, `Contains` has to iterate through the array whereas the hash set can perform a hash lookup.

However, as it's unlikely to make any difference in our case feel free to leave as it is – it's up to you.

tests/NHSD.BuyingCatalogue.Organisations.Api.UnitTests/Repository/OdsRepositoryTests.cs
assoomro: It will be a few lines, but I think you're right and it will be worth it.

src/NHSD.BuyingCatalogue.Organisations.Api/Repositories/OdsRepository.cs
assoomro: I was not aware, thanks for pointing it out! I much prefer your version.

src/NHSD.BuyingCatalogue.Organisations.Api/Settings/OdsSettings.cs
assoomro: I'm all up for small optimisations that can help, however in this instance, as the search space is and will be very narrow - [maximum 5 elements](https://digital.nhs.uk/services/organisation-data-service/guidance-for-developers/search-parameters#primaryroleid-and-nonprimaryroleid)  I believe such change would make no difference, according to [this](https://stackoverflow.com/questions/150750/hashset-vs-list-performance), it would actually be slower! (I'm assuming that operations on arrays are even faster than on List<T>s.)

PR #151, opened 2020-04-21
3536 create organisation post endpoint

assoomro:


src/NHSD.BuyingCatalogue.Organisations.Api/Controllers/OrganisationsController.cs
```diff
@@ -114,22 +118,36 @@ public async Task<ActionResult> UpdateOrganisationByIdAsync(Guid id, UpdateOrgan
...
                 throw new ArgumentNullException(nameof(viewModel));
-            }
 
-            //TODO: Task #6168
+            var result = await _createOrganisationService.CreateAsync(new CreateOrganisationRequest(
+                viewModel.OrganisationName,
+                viewModel.OdsCode,
+                viewModel.PrimaryRoleId,
+                viewModel.CatalogueAgreementSigned,
+                viewModel.Address is null ? null : new Address()
```
bjss-chris, line 128: The brackets `()` can be omitted when using object initialization.

src/NHSD.BuyingCatalogue.Organisations.Api/Controllers/OrganisationsController.cs
```diff
@@ -114,22 +118,36 @@ public async Task<ActionResult> UpdateOrganisationByIdAsync(Guid id, UpdateOrgan
...
+                    Town = viewModel.Address.Town,
+                    County = viewModel.Address.County,
+                    Postcode = viewModel.Address.Postcode,
+                    Country = viewModel.Address.Country
+                }
+            ));
 
-            // Canned data
-            var firstOrganisation = (await _organisationRepository.ListOrganisationsAsync()).First();
+            var response = new CreateOrganisationResponseViewModel();
```
bjss-chris, line 141: There are no changes necessary now but I'll be glad when we fix the naming of these `ViewModel` classes that aren't view models. I had to do a double-take because I thought we were MVC here and returning a view but we're not – it's API.

src/NHSD.BuyingCatalogue.Organisations.Api/Services/CreateOrganisationRequest.cs
```diff
@@ -0,0 +1,26 @@
...
+
+        public string OdsCode { get; }
+
+        public string PrimaryRoleId { get; }
+
+        public bool CatalogueAgreementSigned { get; }
+
+        public Address Address { get; }
+
+        public CreateOrganisationRequest(string organisationName, string odsCode, string primaryRoleId, bool catalogueAgreementSigned, Address address)
```
bjss-chris, line 17: There's a few too many constructor parameters here for my liking but I understand why you've taken this approach. However, I'm not sure the class really needs to be immutable...

src/NHSD.BuyingCatalogue.Organisations.Api/Services/ICreateOrganisationService.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System.Threading.Tasks;
+using NHSD.BuyingCatalogue.Identity.Common.Results;
+
+namespace NHSD.BuyingCatalogue.Organisations.Api.Services
+{
+    public interface ICreateOrganisationService
+    {
+        Task<Result<string>> CreateAsync(CreateOrganisationRequest request);
```
bjss-chris, line 8: Can we return `Result<Guid>` here?

src/NHSD.BuyingCatalogue.Organisations.Api/Services/CreateOrganisationService.cs
```diff
@@ -0,0 +1,47 @@
...
+            };
+
+            var validationResult = await _organisationValidator.ValidateAsync(organisation);
+
+            if (!validationResult.IsSuccess)
+                return Result.Failure<string>(validationResult.Errors);
+
+            await _organisationRepository.CreateOrganisationAsync(organisation);
+
+            return Result.Success(organisation.OrganisationId.ToString());
```
bjss-chris, line 44: Can we use the `Guid` here rather than doing `ToString()`?

tests/NHSD.BuyingCatalogue.Organisations.Api.UnitTests/Builders/OrganisationControllerBuilder.cs
```diff
@@ -48,15 +53,32 @@ internal OrganisationControllerBuilder WithUpdateOrganisation(Organisation organ
             return this;
         }
 
+        internal OrganisationControllerBuilder WithCreateOrganisation()
+        {
+            var repositoryMock = new Mock<IOrganisationRepository>();
+            repositoryMock.Setup(x => x.CreateOrganisationAsync(It.IsAny<Organisation>()));
+            _organisationRepository = repositoryMock.Object;
+            return this;
+        }
```
bjss-chris, line 62: Is this ever used/required without `WithCreateOrganisationService`?

tests/NHSD.BuyingCatalogue.Organisations.Api.UnitTests/Builders/OrganisationControllerBuilder.cs
```diff
@@ -48,15 +53,32 @@ internal OrganisationControllerBuilder WithUpdateOrganisation(Organisation organ
...
+
         public OrganisationControllerBuilder WithOrganisationRepository(IOrganisationRepository organisationRepository)
         {
             _organisationRepository = organisationRepository;
             return this;
         }
 
+        public OrganisationControllerBuilder WithCreateOrganisationService(bool isSuccess, string resultValue)
```
bjss-chris, line 70: I think this could probably be named better... Perhaps `WithCreateOrganisationResult` or something along those lines?

tests/NHSD.BuyingCatalogue.Organisations.Api.UnitTests/Controllers/OrganisationControllerTests.cs
```diff
@@ -19,6 +22,24 @@ public sealed class OrganisationControllerTests
     {
         private readonly Address _address1 = AddressBuilder.Create().WithLine1("18 Stone Road").Build();
 
+        [Test]
+        public void Constructor_Null_Repository_Throws()
```
bjss-chris, line 26: I think `_NullRepository_` would be clearer.

tests/NHSD.BuyingCatalogue.Organisations.Api.UnitTests/Controllers/OrganisationControllerTests.cs
```diff
@@ -19,6 +22,24 @@ public sealed class OrganisationControllerTests
...
+        {
+            Assert.Throws<ArgumentNullException>(() =>
+            {
+                var _ = new OrganisationsController(null, Mock.Of<ICreateOrganisationService>());
+            });
+        }
+
+        [Test]
+        public void Constructor_Null_Service_Throws()
```
bjss-chris, line 35: I think `_NullService_` would be clearer.

tests/NHSD.BuyingCatalogue.Organisations.Api.UnitTests/Controllers/OrganisationControllerTests.cs
```diff
@@ -263,11 +284,61 @@ public async Task UpdateOrganisationByIdAsync_OrganisationRepository_UpdateAsync
...
+
+            actual.Should().BeEquivalentTo(expected);
+        }
+
+        [Test]
+        public void CreateOrganisationAsync_NullCreateOrganisationRequestViewModel_Throws()
+        {
+            Assert.ThrowsAsync<ArgumentNullException>(async () =>
+            {
+                using var controller = OrganisationControllerBuilder.Create().WithCreateOrganisation().Build();
```
bjss-chris, line 339: Is `WithCreateOrganisation()` required here?

tests/NHSD.BuyingCatalogue.Organisations.Api.UnitTests/Services/CreateOrganisationServiceTests.cs
```diff
@@ -0,0 +1,153 @@
...
+using NUnit.Framework;
+
+namespace NHSD.BuyingCatalogue.Organisations.Api.UnitTests.Services
+{
+    [TestFixture]
+    [Parallelizable(ParallelScope.All)]
+    public sealed class CreateOrganisationServiceTests
+    {
+        [Test]
+        public void Constructor_Null_Repository_Throws()
```
bjss-chris, line 22: I think `_NullRepository_` would be clearer.

tests/NHSD.BuyingCatalogue.Organisations.Api.UnitTests/Services/CreateOrganisationServiceTests.cs
```diff
@@ -0,0 +1,153 @@
...
+        {
+            Assert.Throws<ArgumentNullException>(() =>
+            {
+                var _ = new CreateOrganisationService(null, Mock.Of<IOrganisationValidator>());
+            });
+        }
+
+        [Test]
+        public void Constructor_Null_Validator_Throws()
```
bjss-chris, line 31: I think `_NullValidator_` would be clearer.

tests/NHSD.BuyingCatalogue.Organisations.Api.UnitTests/Services/CreateOrganisationServiceTests.cs
```diff
@@ -0,0 +1,153 @@
...
+        {
+            Assert.Throws<ArgumentNullException>(() =>
+            {
+                var _ = new CreateOrganisationService(Mock.Of<IOrganisationRepository>(), null);
+            });
+        }
+
+        [Test]
+        public void CreateAsync_Null_Request_Throws()
```
bjss-chris, line 40: I think `_NullRequest_` would be clearer.

tests/NHSD.BuyingCatalogue.Organisations.Api.UnitTests/Services/CreateOrganisationServiceTests.cs
```diff
@@ -0,0 +1,153 @@
...
+                expected.Name,
+                expected.OdsCode,
+                expected.PrimaryRoleId,
+                expected.CatalogueAgreementSigned,
+                expected.Address));
+
+            calledBack.Should().BeTrue();
+        }
+
+        public static IOrganisationRepository SetUpRepository()
```
bjss-chris, line 123: Can this be `private`?

tests/NHSD.BuyingCatalogue.Organisations.Api.UnitTests/Services/CreateOrganisationServiceTests.cs
```diff
@@ -0,0 +1,153 @@
...
+        {
+            return SetUpRepositoryMock().Object;
+        }
+
+        public static Mock<IOrganisationRepository> SetUpRepositoryMock()
```
bjss-chris, line 128: Can this be `private`?

tests/NHSD.BuyingCatalogue.Organisations.Api.UnitTests/Services/CreateOrganisationServiceTests.cs
```diff
@@ -0,0 +1,153 @@
...
+using NHSD.BuyingCatalogue.Organisations.Api.Services;
+using NHSD.BuyingCatalogue.Organisations.Api.UnitTests.Builders;
+using NHSD.BuyingCatalogue.Organisations.Api.Validators;
+using NUnit.Framework;
+
+namespace NHSD.BuyingCatalogue.Organisations.Api.UnitTests.Services
+{
+    [TestFixture]
+    [Parallelizable(ParallelScope.All)]
+    public sealed class CreateOrganisationServiceTests
```
bjss-chris, line 19: Not important but can be `internal`.

tests/NHSD.BuyingCatalogue.Organisations.Api.UnitTests/Validators/OrganisationValidatorTests.cs
```diff
@@ -0,0 +1,91 @@
...
+using NHSD.BuyingCatalogue.Organisations.Api.Repositories;
+using NHSD.BuyingCatalogue.Organisations.Api.UnitTests.Builders;
+using NHSD.BuyingCatalogue.Organisations.Api.Validators;
+using NUnit.Framework;
+
+namespace NHSD.BuyingCatalogue.Organisations.Api.UnitTests.Validators
+{
+    [TestFixture]
+    [Parallelizable(ParallelScope.All)]
+    public sealed class OrganisationValidatorTests
```
bjss-chris, line 16: Not important but can be `internal`.

tests/NHSD.BuyingCatalogue.Organisations.Api.UnitTests/Validators/OrganisationValidatorTests.cs
```diff
@@ -0,0 +1,91 @@
...
+
+            var validator = new OrganisationValidator(mockOrganisationRepository.Object);
+
+            await validator.ValidateAsync(newOrganisation);
+
+            mockOrganisationRepository.Verify(r => r.GetByNameAsync(newOrganisation.Name), Times.Once);
+        }
+
+        [Test]
+        public void Validate_Null_Organisation_Throws()
```
bjss-chris, line 74: I think `_NullOrganisation_` would be clearer.

src/NHSD.BuyingCatalogue.Organisations.Api/Controllers/OrganisationsController.cs
assoomro: 👍

src/NHSD.BuyingCatalogue.Organisations.Api/Controllers/OrganisationsController.cs
assoomro: Agreed

tests/NHSD.BuyingCatalogue.Organisations.Api.UnitTests/Controllers/OrganisationControllerTests.cs
assoomro: 👍

tests/NHSD.BuyingCatalogue.Organisations.Api.UnitTests/Controllers/OrganisationControllerTests.cs
assoomro: 👍

tests/NHSD.BuyingCatalogue.Organisations.Api.UnitTests/Services/CreateOrganisationServiceTests.cs
assoomro: 👍

tests/NHSD.BuyingCatalogue.Organisations.Api.UnitTests/Services/CreateOrganisationServiceTests.cs
assoomro: 👍

tests/NHSD.BuyingCatalogue.Organisations.Api.UnitTests/Services/CreateOrganisationServiceTests.cs
assoomro: 👍

tests/NHSD.BuyingCatalogue.Organisations.Api.UnitTests/Services/CreateOrganisationServiceTests.cs
assoomro: 👍

tests/NHSD.BuyingCatalogue.Organisations.Api.UnitTests/Services/CreateOrganisationServiceTests.cs
assoomro: 👍

tests/NHSD.BuyingCatalogue.Organisations.Api.UnitTests/Services/CreateOrganisationServiceTests.cs
assoomro: 👍

tests/NHSD.BuyingCatalogue.Organisations.Api.UnitTests/Validators/OrganisationValidatorTests.cs
assoomro: 👍

tests/NHSD.BuyingCatalogue.Organisations.Api.UnitTests/Validators/OrganisationValidatorTests.cs
assoomro: 👍

tests/NHSD.BuyingCatalogue.Organisations.Api.UnitTests/Builders/OrganisationControllerBuilder.cs
assoomro: Not really.

Would you suggest refactoring the code in this method to be part of `WithCreateOrganisationService`  (soon to be renamed) instead?

src/NHSD.BuyingCatalogue.Organisations.Api/Services/CreateOrganisationRequest.cs
assoomro: I like it being immutable. On the other hand, I get your point, it might be a bit of a pain only being to instantiate this class using the constructor, however, given the purpose of this class, I think immutability suits it well.

tests/NHSD.BuyingCatalogue.Organisations.Api.UnitTests/Builders/OrganisationControllerBuilder.cs
assoomro: `WithCreateOrganisationResult` sounds like an improvement.
How about `WithCreateOrganisationReturningResult`

bjss-chris: `WithCreateOrganisationReturningResult` sounds good to me.

tests/NHSD.BuyingCatalogue.Organisations.Api.UnitTests/Builders/OrganisationControllerBuilder.cs
bjss-chris: I think I'd be tempted to do that or make it a private method that you invoke from the renamed `WithCreateOrganisationService`.

assoomro: 👍

tests/NHSD.BuyingCatalogue.Organisations.Api.UnitTests/Builders/OrganisationControllerBuilder.cs
assoomro: 👍

src/NHSD.BuyingCatalogue.Organisations.Api/Services/ICreateOrganisationService.cs
assoomro: Interesting point, thank you for raising it.

I'd prefer the service to return a string representing the guid - as that's what gets propagated to the (wrongly named) viewModel.

Why doesnt the 'viewModel' return the Guid rather than string?
 I chose the current implementation because it felt right. I know that's not a strong argument, however I felt wrong about returning a Guid without having a concrete understanding of what/where and how serialises that Guid to JSON (which is not a concern with a string).

I had a look and the rework to return a Guid all the way through is not huge (6 files).

Conclusion, I'm not a fan, however, if you shed some more light on why should we return a Guid instead of a string I'm willing to compromise and change it.

src/NHSD.BuyingCatalogue.Organisations.Api/Services/CreateOrganisationService.cs
assoomro: Will resolve this based on the outcomes of comments above.

src/NHSD.BuyingCatalogue.Organisations.Api/Services/ICreateOrganisationService.cs
bjss-chris: `CreateOrganisationResponseViewModel.OrganisationId` can be a `Guid` – it will be serialized as a string. Sorry, I probably just have picked that up and left a comment.

The reason to prefer a `Guid` is type-safety and accurate object modelling. In the OO world we often have a tendency to be a bit lazy with our typing, such as using `string` for URLs when we should be using `Uri`. There are of course plenty of other examples too, such as using `string` for e-mails etc. – no doubt you'll be able to think of many more examples. Functional programmers tend to be much better at defining specific types... 

When we go for the `string` option we lose the advantages that the strong-typing brings – such as making sure we always have a valid Guid as our ID. It also means we can take advantage of any useful operations that the Guid struct provides (although we may not need them right now).

Therefore, I would still recommend refactoring to use the `Guid` type – particularly as it's already available in the framework.

P.S. I wasn't at all suggesting that you've been lazy in your implementation – anything but as you've obviously given it some thought! 👍

assoomro: Alright I'm sold. Guid it is 😄

src/NHSD.BuyingCatalogue.Organisations.Api/Services/CreateOrganisationService.cs
assoomro: 👍

tests/NHSD.BuyingCatalogue.Organisations.Api.UnitTests/Builders/OrganisationControllerBuilder.cs
assoomro: On second thought, wouldn't `WithCreateOrganisationServiceReturningResult` be even better? Given that we're setting up the _service_ and not the organisation to return a result.

tests/NHSD.BuyingCatalogue.Organisations.Api.UnitTests/Builders/OrganisationControllerBuilder.cs
```diff
@@ -53,29 +53,29 @@ internal OrganisationControllerBuilder WithUpdateOrganisation(Organisation organ
...
+        public OrganisationControllerBuilder WithCreateOrganisationServiceReturningResult(bool isSuccess, string resultValue)
         {
+            WithCreateOrganisation();
             var createOrganisationService = new Mock<ICreateOrganisationService>();
             createOrganisationService.Setup(s => s.CreateAsync(It.IsAny<CreateOrganisationRequest>()))
-                .ReturnsAsync(isSuccess ? Result.Success(resultValue) : Result.Failure<string>(new ErrorDetails(resultValue)));
+                .ReturnsAsync(isSuccess ? Result.Success(Guid.Parse(resultValue)) : Result.Failure<Guid>(new ErrorDetails(resultValue)));
             _createOrganisationService = createOrganisationService.Object;
             return this;
         }
```
bjss-chris: I was just thinking about this and I think I'd be inclined to refactor this slightly as I think it's best to avoid logic in tests and test setup methods wherever possible. I think I'd probably do something like this:

```c#
internal OrganisationControllerBuilder WithCreateOrganisationServiceReturningResult(Guid result)
{
	return WithCreateOrganisationServiceReturningResult(Result.Success(result));
}

internal OrganisationControllerBuilder WithCreateOrganisationServiceReturningResult(string result)
{
	return WithCreateOrganisationServiceReturningResult(Result.Failure<Guid>(new ErrorDetails(result)));
}

private OrganisationControllerBuilder WithCreateOrganisationServiceReturningResult(Result<Guid> result)
{
	WithCreateOrganisation();
	var createOrganisationService = new Mock<ICreateOrganisationService>();
	createOrganisationService.Setup(s => s.CreateAsync(It.IsAny<CreateOrganisationRequest>()))
		.ReturnsAsync(result);
	_createOrganisationService = createOrganisationService.Object;
	return this;
}
```
What do you think?

(Again, it's not that important but the `With` methods you've added can all be `internal`.)

assoomro: Love it. 

I tend to have the habit of leaving things more open - leaving users options if you will. 
After a conversation with @sam-roberts-bjss on similar topic, I understand that this not preferable in our setting.

I will try to keep in this mind going forward, thank you for revisiting this!

src/NHSD.BuyingCatalogue.Organisations.Api/Validators/OrganisationValidator.cs
```diff
@@ -0,0 +1,32 @@
...
+        public async Task<Result> ValidateAsync(Organisation organisation)
+        {
+            if (organisation is null)
+                throw new ArgumentNullException(nameof(organisation));
+
+            var persistedOrganisation = await _organisationRepository.GetByOdsCodeAsync(organisation.OdsCode);
+
+            return persistedOrganisation is null
+                ? Result.Success()
+                : Result.Failure(new List<ErrorDetails> {new ErrorDetails("OrganisationAlreadyExists")});
```
sam-roberts-bjss, line 29: Although it is only one error at the moment, we might add more in the future. To encapsulate all the errors in one place for the organisation I would create an OrganisationErrors static class e.g. ApplicationUserErrors. This will allow re-use within the unit tests & encapsulate any parameters the error details required. However in this case there are no parameters.

tests/NHSD.BuyingCatalogue.Identity.Api.IntegrationTests/Steps/OrganisationsSteps.cs
```diff
@@ -123,6 +124,22 @@ public async Task WhenAPUTRequestIsMadeForAnOrganisationWithNameOrganisation(str
...
+            var data = table.CreateInstance<OrganisationTable>();
+
+            using var client = new HttpClient();
+            client.SetBearerToken(_context.Get(ScenarioContextKeys.AccessToken, ""));
+
+            var payload = JsonConvert.SerializeObject(TransformOrganisationIntoPayload(data));
+            using var content = new StringContent(payload, Encoding.UTF8, "application/json");
+            _response.Result = await client.PostAsync(new Uri($"{_organisationUrl}"), content);
+
+            if (_response.Result.StatusCode == HttpStatusCode.Created)
```
sam-roberts-bjss, line 139: It probably doesn't matter but I was thinking for consistency reasons we would try to use the `Microsoft.AspNetCore.Http.StatusCodes` instead of the `System.Net` ones e.g. `Status201Created`

tests/NHSD.BuyingCatalogue.Organisations.Api.UnitTests/Validators/OrganisationValidatorTests.cs
```diff
@@ -0,0 +1,91 @@
...
+        {
+            var newOrganisation = OrganisationBuilder.Create(1).Build();
+
+            var mockOrganisationRepository = SetUpGetByOdsCodeAsync(null);
+
+            var validator = new OrganisationValidator(mockOrganisationRepository);
+
+            var result = await validator.ValidateAsync(newOrganisation);
+
+            result.IsSuccess.Should().BeTrue();
```
sam-roberts-bjss, line 38: It makes the code easier to read and understand when there is just one assert function per test. This suggestion could be applied here e.g. actual.Should().Be(Result.Success());

This also highlights the use of the 'actual' variable, that I believe we said in the coding standards meeting we would introduce (along with 'expected' but probably not worth it here).

tests/NHSD.BuyingCatalogue.Organisations.Api.UnitTests/Builders/OrganisationControllerBuilder.cs
```diff
@@ -53,18 +53,28 @@ internal OrganisationControllerBuilder WithUpdateOrganisation(Organisation organ
...
+        {
+            return WithCreateOrganisationServiceReturningResult(Result.Success(result));
+        }
+
+        internal OrganisationControllerBuilder WithCreateOrganisationServiceReturningFailure(string result)
+        {
+            return WithCreateOrganisationServiceReturningResult(Result.Failure<Guid>(new ErrorDetails(result)));
+        }
+
+        internal OrganisationControllerBuilder WithCreateOrganisationServiceReturningResult(Result<Guid> result)
```
bjss-chris, line 72: I'd probably make this `private` to encourage the use of `WithCreateOrganisationServiceReturningSuccess` and `WithCreateOrganisationServiceReturningFailure`. Do you think it might be useful exposing it?

src/NHSD.BuyingCatalogue.Organisations.Api/Controllers/OrganisationsController.cs
```diff
@@ -114,22 +118,36 @@ public async Task<ActionResult> UpdateOrganisationByIdAsync(Guid id, UpdateOrgan
...
                 throw new ArgumentNullException(nameof(viewModel));
-            }
 
-            //TODO: Task #6168
+            var result = await _createOrganisationService.CreateAsync(new CreateOrganisationRequest(
+                viewModel.OrganisationName,
+                viewModel.OdsCode,
+                viewModel.PrimaryRoleId,
+                viewModel.CatalogueAgreementSigned,
+                viewModel.Address is null ? null : new Address
```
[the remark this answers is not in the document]
assoomro: 👍 Good suggestion.

src/NHSD.BuyingCatalogue.Organisations.Api/Controllers/OrganisationsController.cs
```diff
@@ -114,22 +118,36 @@ public async Task<ActionResult> UpdateOrganisationByIdAsync(Guid id, UpdateOrgan
...
-                Errors = null,
-                OrganisationId = firstOrganisation.OrganisationId.ToString()
-            };
+                response.Errors = result.Errors.Select(x => new ErrorMessageViewModel(x.Id, x.Field));
+                return BadRequest(response);
+            }
 
-            return Created(new Uri($"/{response.OrganisationId}", UriKind.Relative), response);
+            response.OrganisationId = result.Value;
+            return Created(new Uri($"/{result.Value}", UriKind.Relative), response);
```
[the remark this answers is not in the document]
assoomro: 👍 I'm assuming we're leaving it as is and fixing it as part of [6259](https://buyingcatalog.visualstudio.com/Buying%20Catalogue/_sprints/taskboard/Buying%20Catalogue%20Tech%20Cell/Buying%20Catalogue/Iteration%2023) right?

src/NHSD.BuyingCatalogue.Organisations.Api/Validators/OrganisationValidator.cs
assoomro: 👍

src/NHSD.BuyingCatalogue.Organisations.Api/Repositories/OrganisationRepository.cs
```diff
@@ -27,6 +27,20 @@ public async Task<Organisation> GetByIdAsync(Guid id)
             return await _context.Organisations.FirstOrDefaultAsync(org => org.OrganisationId == id);
         }
 
+        public async Task<Organisation> GetByOdsCodeAsync(string odsCode)
+        {
+            return await _context.Organisations.FirstOrDefaultAsync(org => org.OdsCode == odsCode);
```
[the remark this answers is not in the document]
bjss-chris: Case will only matter in the SQL if the database is configured with a case-sensitive collation (it's not at the moment as we've no requirement for it).

However, I guess whether we tweak this at all depends on if we want to 'belt-and-brace' it anyway...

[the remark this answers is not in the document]
bjss-chris: The casing in the SQL would only matter if we configured the database using a case-sensitive collation. As we're not, we'd only need to tweak this if we want to belt-and-brace it anyway (as it would probably just involve adding a `ToUpperInvariant()`...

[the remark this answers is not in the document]
assoomro: After some research I came across [this](https://stackoverflow.com/a/28391284/12343768) claiming that the default collation is case insensitive. Being the sceptic I am I went over to SSMS and tried it out. And to my surprise, the stranger on the internet was right. 😄 

Lower:
image


Upper:
image

To conclude, given our default settings, this should not be an issue.

tests/NHSD.BuyingCatalogue.Identity.Api.IntegrationTests/Steps/OrganisationsSteps.cs
assoomro: 👍 Consistency rules

tests/NHSD.BuyingCatalogue.Organisations.Api.UnitTests/Validators/OrganisationValidatorTests.cs
assoomro: 👍 Agreed & did something similar with the failure test.

tests/NHSD.BuyingCatalogue.Organisations.Api.UnitTests/Builders/OrganisationControllerBuilder.cs
assoomro: Typo (of sorts), it should be private here. Leaving it internal will be confusing.

src/NHSD.BuyingCatalogue.Organisations.Api/Models/OrganisationErrors.cs
```diff
@@ -0,0 +1,12 @@
+﻿using NHSD.BuyingCatalogue.Identity.Common.Models;
+
+namespace NHSD.BuyingCatalogue.Organisations.Api.Models
+{
+    public static class OrganisationErrors
+    {
+        public static ErrorDetails OrganisationAlreadyExists()
```
bjss-chris, line 7: Not that important but these can both (class and method) be `internal`.

assoomro: I'm thinking about leaving them public for now. if I were to change them to be internal, I'd also need to change the ones in `ApplicationUserErrors` and I don't want to bloat this PR more than it needs to be. Hope that's understandable.

bjss-chris: Why would `ApplicationUserErrors` need to change? I didn't think that was part of this PR?

Are you just thinking in terms of consistency? If so, we could change `ApplicationUserErrors` later. However, as above it's not that important.

assoomro: 😄 Yeah I meant for consistency sake.

So are we ok with leaving it as is for now?

bjss-chris: I don't mind. If we end up making any further changes I'd suggest changing it; otherwise, just leave it as is for now.

assoomro: Sounds good to me.

I'll wait and see whether there's anything else to amend after @sam-roberts-bjss re-reviews this.

If yes, I'll change it in that change set, if not, I'll leave it as is and merge.

src/NHSD.BuyingCatalogue.Organisations.Api/Controllers/OrganisationsController.cs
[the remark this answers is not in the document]
sam-roberts-bjss: I mean I would prefer if we fixed it now so we are not introducing more tech debt & will mitigate any "copying" of the functionality. If you feel pressed to tackle this change as part of the other task then OK, but it might be a good another just to add a note to the other task 6259 to indicate the additional fix needed.

src/NHSD.BuyingCatalogue.Organisations.Api/Repositories/OrganisationRepository.cs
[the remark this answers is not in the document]
sam-roberts-bjss: 👍

tests/NHSD.BuyingCatalogue.Identity.Api.IntegrationTests/Steps/OrganisationsSteps.cs
sam-roberts-bjss: Wow that is a shame. Yeah OK, apologies I didn't think it would that much trouble.

src/NHSD.BuyingCatalogue.Organisations.Api/Controllers/OrganisationsController.cs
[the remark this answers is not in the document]
assoomro: 👍 I will investigate this task as part of this. If I don't get anywhere by the end of today I'm going to merge as is - returning a 201 with the awful /{id} location header.

src/NHSD.BuyingCatalogue.Identity.Common/Extensions/StringExtensions.cs
```diff
@@ -0,0 +1,15 @@
...
+namespace NHSD.BuyingCatalogue.Identity.Common.Extensions
+{
+    public static class StringExtensions
+    {
+        public static string TrimAsync(this string input)
+        {
+            if (input is null)
+                throw new ArgumentNullException(nameof(input));
+
+            return input.Replace("async", "", StringComparison.OrdinalIgnoreCase);
```
sam-roberts-bjss, line 12: Rather than replace, shouldn't this just be stripping the Async off the end of the method name? I think there is an example in the `PasswordResetCallback` class, `Controller` method. Also I would be tempted to not throw an exception if the `input` is `null`. This means the method calling this extension doesn't have to check every time. If the input value is null just return null.

tests/NHSD.BuyingCatalogue.Identity.Common.UnitTests/Extensions/StringExtensionsTests.cs
```diff
@@ -0,0 +1,51 @@
...
+
+namespace NHSD.BuyingCatalogue.Identity.Common.UnitTests.Extensions
+{
+    [TestFixture]
+    [Parallelizable(ParallelScope.All)]
+    internal sealed class StringExtensionsTests
+    {
+
+        [Test]
+        public void TrimAsync__NullInput_Throws()
```
sam-roberts-bjss, line 14: Any reason for the double underscore?

src/NHSD.BuyingCatalogue.Identity.Common/Extensions/StringExtensions.cs
assoomro: As per our conversation, I agree with your point. After spending quite some time on making the compiler behave, I gave up and went with throwing if input is null. Given the use cases for this, it's not harmful to throw at that point (we would throw anyway, just bit later).

tests/NHSD.BuyingCatalogue.Identity.Common.UnitTests/Extensions/StringExtensionsTests.cs
assoomro: 👍 

A typo 😅

src/NHSD.BuyingCatalogue.Identity.Common/Extensions/StringExtensions.cs
sam-roberts-bjss: 👍

PR #152, opened 2020-04-21
Add user agreement page and corresponding logic

bjss-chris:
There are some outstanding tasks still required, particularly around the styling of the consent view and its content, so I have added the following tasks to [story 3924](https://dev.azure.com/BuyingCatalog/Buying%20Catalogue/_workitems/edit/3924):

[API/FE – Correct the styling of the consent view (6331)](https://dev.azure.com/BuyingCatalog/Buying%20Catalogue/_workitems/edit/6331)
[API – Confirm correct validation error message (6329)](https://dev.azure.com/BuyingCatalog/Buying%20Catalogue/_workitems/edit/6329)
[API – Confirm correct agreement text re privacy (6330)](https://dev.azure.com/BuyingCatalog/Buying%20Catalogue/_workitems/edit/6330)
[API – Add integration tests (6332)](https://dev.azure.com/BuyingCatalog/Buying%20Catalogue/_workitems/edit/6332)
[API – Move agreement text to DB (6333)](https://dev.azure.com/BuyingCatalog/Buying%20Catalogue/_workitems/edit/6333)

AB#6135
AB#6173

### DIFF
--- a/src/NHSD.BuyingCatalogue.Identity.Api/Repositories/ScopeRepository.cs
+++ b/src/NHSD.BuyingCatalogue.Identity.Api/Repositories/ScopeRepository.cs
@@ -1,0 +1,29 @@
+﻿using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+using System.Linq;
+using NHSD.BuyingCatalogue.Identity.Api.Settings;
+
+namespace NHSD.BuyingCatalogue.Identity.Api.Repositories
+{
+    internal sealed class ScopeRepository : IScopeRepository
+    {
+        private readonly List<string> _scopes = new List<string>();
+
+        [SuppressMessage(
+            "Globalization",
+            "CA1308:Normalize strings to uppercase",
+            Justification = "OpenId scopes are lower case and must match")]
+        public ScopeRepository(
+            IEnumerable<ResourceSetting> apiResources,
+            IEnumerable<IdentityResourceSetting> identityResources)
+        {
+            if (apiResources != null)
+                _scopes.AddRange(apiResources.Select(r => r.ResourceName));
+
+            if (identityResources != null)
+                _scopes.AddRange(identityResources.Select(r => r.ResourceType.ToLowerInvariant()));
+        }
+
+        public IEnumerable<string> Scopes => _scopes;
+    }
+}
--- a/src/NHSD.BuyingCatalogue.Identity.Api/Repositories/ScopeRepository.cs
+++ b/src/NHSD.BuyingCatalogue.Identity.Api/Repositories/ScopeRepository.cs
@@ -24,6 +24,6 @@ namespace NHSD.BuyingCatalogue.Identity.Api.Repositories
                 _scopes.AddRange(identityResources.Select(r => r.ResourceType.ToLowerInvariant()));
         }
 
-        public IEnumerable<string> Scopes => _scopes;
+        public IReadOnlyCollection<string> Scopes => _scopes;
     }
 }
--- a/src/NHSD.BuyingCatalogue.Identity.Api/Services/AgreementConsentService.cs
+++ b/src/NHSD.BuyingCatalogue.Identity.Api/Services/AgreementConsentService.cs
@@ -38,7 +38,7 @@ namespace NHSD.BuyingCatalogue.Identity.Api.Services
                 throw new ArgumentNullException(nameof(returnUrl));
 
             var context = await _interaction.GetAuthorizationContextAsync(returnUrl.ToString());
-            if (context == null)
+            if (context is null)
                 return Result.Failure<Uri>();
 
             var consent = new ConsentResponse
--- a/src/NHSD.BuyingCatalogue.Identity.Api/Services/AgreementConsentService.cs
+++ b/src/NHSD.BuyingCatalogue.Identity.Api/Services/AgreementConsentService.cs
@@ -1,0 +1,61 @@
+﻿using System;
+using System.Threading.Tasks;
+using IdentityServer4.Events;
+using IdentityServer4.Models;
+using IdentityServer4.Services;
+using NHSD.BuyingCatalogue.Identity.Api.Repositories;
+using NHSD.BuyingCatalogue.Identity.Common.Results;
+
+namespace NHSD.BuyingCatalogue.Identity.Api.Services
+{
+    internal sealed class AgreementConsentService : IAgreementConsentService
+    {
+        private readonly IEventService _eventService;
+        private readonly IIdentityServerInteractionService _interaction;
+        private readonly IScopeRepository _scopeRepository;
+
+        public AgreementConsentService(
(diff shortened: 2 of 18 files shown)


src/NHSD.BuyingCatalogue.Identity.Api/Repositories/ScopeRepository.cs
```diff
@@ -0,0 +1,29 @@
...
+            IEnumerable<IdentityResourceSetting> identityResources)
+        {
+            if (apiResources != null)
+                _scopes.AddRange(apiResources.Select(r => r.ResourceName));
+
+            if (identityResources != null)
+                _scopes.AddRange(identityResources.Select(r => r.ResourceType.ToLowerInvariant()));
+        }
+
+        public IEnumerable<string> Scopes => _scopes;
```
sam-roberts-bjss, line 27: One suggestion here is to return a readonly collection rather than the `IEnumerable` as this will state explicit intent. This only applies if we decide the list of scopes should not be added to once initialised, which I don't think we do.

src/NHSD.BuyingCatalogue.Identity.Api/Services/AgreementConsentService.cs
```diff
@@ -0,0 +1,61 @@
...
+            return await _interaction.GetAuthorizationContextAsync(returnUrl.ToString()) != null;
+        }
+
+        public async Task<Result<Uri>> GrantConsent(Uri returnUrl, string subjectId)
+        {
+            if (returnUrl is null)
+                throw new ArgumentNullException(nameof(returnUrl));
+
+            var context = await _interaction.GetAuthorizationContextAsync(returnUrl.ToString());
+            if (context == null)
```
sam-roberts-bjss: Very minor -> For consistency reasons the use of `== null` is different to how the `returnUrl` is checked.

src/NHSD.BuyingCatalogue.Identity.Api/Services/AgreementConsentService.cs
```diff
@@ -0,0 +1,61 @@
...
+                return Result.Failure<Uri>();
```
sam-roberts-bjss, line 42: Unexpected behaviour around returning a failure but not expressing why i.e. no error detail passed into the Result<T> class. Though what you would tell the user of this method I am not sure.

src/NHSD.BuyingCatalogue.Identity.Api/Startup.cs
```diff
@@ -76,13 +77,15 @@ public void ConfigureServices(IServiceCollection services)
             services.AddSingleton(cookieExpiration);
             services.AddSingleton(disabledErrorMessage);
             services.AddSingleton(registrationSettings);
+            services.AddSingleton<IScopeRepository>(new ScopeRepository(apiResources, identityResources));
```
sam-roberts-bjss, line 80: It might be a slight tidy to the code (slight as it is probably adding more code though) & help in the future if more dependencies get added to the `ScopeRepository`, but the suggestion here would be to register the `ApiResourceSettingCollection` & the `IdentityResourceSettingCollection` with the service collection & use dependency injection to create the `ScopeRepository`.

src/NHSD.BuyingCatalogue.Identity.Api/ViewModels/Consent/ConsentViewModel.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System;
+using System.ComponentModel;
+using System.ComponentModel.DataAnnotations;
+
+namespace NHSD.BuyingCatalogue.Identity.Api.ViewModels.Consent
+{
+    public sealed class ConsentViewModel
+    {
+        // TODO: confirm correct error message
```
sam-roberts-bjss, line 9: Would you mind adding the task associated with this TODO message?

src/NHSD.BuyingCatalogue.Identity.Api/Services/AgreementConsentService.cs
bjss-chris: I spotted this myself and I think I've already fixed this – but your review and my last commit might have crossed paths so to speak. Sorry, probably should have let you know.

sam-roberts-bjss: Yeah good spot. You weren't to know I was actually doing the review so not to worry.

src/NHSD.BuyingCatalogue.Identity.Api/Startup.cs
bjss-chris: I did actually consider this approach – in fact, it was my initial approach but I decided that (it was more verbose as you've commented) and that we wouldn't really gain anything from it at this point in time. As you say it would definitely be worth tweaking if we add more dependencies to the `ScopeRepository` but possible not right now...

What do you think? I'm happy with either approach.

src/NHSD.BuyingCatalogue.Identity.Api/ViewModels/Consent/ConsentViewModel.cs
bjss-chris: Already done: task 6329 (discussed this morning).

src/NHSD.BuyingCatalogue.Identity.Api/Startup.cs
sam-roberts-bjss: Lets stick with what we have currently. Just wanted to bring it up as a suggestion. Like you said we can revisit if more dependencies are required in the future 👍

src/NHSD.BuyingCatalogue.Identity.Api/ViewModels/Consent/ConsentViewModel.cs
sam-roberts-bjss: Sorry I meant actually adding the task number to the comment in the code. I thought that was what we agreed in the coding standards meeting.

bjss-chris: Sorry, you are of course correct – I should have remembered.

src/NHSD.BuyingCatalogue.Identity.Api/Repositories/ScopeRepository.cs
bjss-chris: I'll refactor to `IReadOnlyCollection<string>`.

src/NHSD.BuyingCatalogue.Identity.Api/Services/AgreementConsentService.cs
bjss-chris: As discussed on our call I'll leave this as is on the basis that it would be difficult to know what to display to the user in the case of an error, and an invalid return URL should only occur in a malicious scenario.

PR #161, opened 2020-04-24
6342 fix location header for 201 responses

assoomro:


assoomro:
> I think we need to review the approach taken here. In essence the changes made go above & beyond what is required in the integration tests.
> 
> The slight concern with using the location response header to retrieve the details of the user is the test has more than one action scenario i.e. it is testing more than one thing. There may be a ripple effect on failed tests if for example there was a problem with the get user endpoint. It is usually better to test each endpoint in isolation.
> 
> All in all I was expecting the assert for the location response header to just check the returned value matches an expected one. What do you think?

My bad 😅 I see your point.

I agree and will revert the changes & add additional step to check the value of the url in the response's location header.

PR #162, opened 2020-04-24
4885 env variable prepend to email

RuthBerry96:


src/NHSD.BuyingCatalogue.Identity.Common/Email/MimeKitExtensions.cs
```diff
@@ -27,7 +28,7 @@ internal static MimeMessage AsMimeMessage(this EmailMessage emailMessage)
 
             var message = new MimeMessage
             {
-                Subject = emailMessage.Subject ?? string.Empty,
+                Subject = $"{emailSubjectPrefix} {emailMessage.Subject}" ?? string.Empty,
```
sam-roberts-bjss, line 31: If the prefix is an empty string we probably do want to trim the whole subject line e.g. Subject = " Test email". Applying a trim will remove the leading space e.g. "Test email".

RuthBerry96: 👍

src/NHSD.BuyingCatalogue.Identity.Common/Email/MimeKitExtensions.cs
```diff
@@ -27,10 +28,13 @@ internal static MimeMessage AsMimeMessage(this EmailMessage emailMessage)
 
             var message = new MimeMessage
             {
-                Subject = emailMessage.Subject ?? string.Empty,
                 Body = bodyBuilder.ToMessageBody(),
             };
 
+            message.Subject = string.IsNullOrWhiteSpace(emailSubjectPrefix) && emailMessage.Subject != null
```
sam-roberts-bjss, line 34: As mentioned on a call. Please review this section of code. The original request was to trim any whitespace from the subject of the email. What happens here if `email.Subject` is equal to an empty string.

RuthBerry96: 👍

PR #170, opened 2020-04-28
6325  ISAPI replaced the Password Hash method with dotnet core implementation

SimonLLawrence:
[AB#6325](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/6325)


SimonLLawrence:
AM#6325

PR #172, opened 2020-04-29
Expired token bugfix

bjss-chris:
Fixes the bug where the reset password view is displayed when an invalid/expired token link is followed instead of immediately redirecting to the expired link view.

[AB#6339](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/6339)


tests/NHSD.BuyingCatalogue.Identity.Api.IntegrationTests/Steps/Common/ScenarioContextKeys.cs
```diff
@@ -2,9 +2,11 @@
 {
     internal static class ScenarioContextKeys
     {
-        public const string AccessToken = "AccessToken";
-        public const string OrganisationMapDictionary = "OrganisationMapDictionary";
-        public const string EmailSent = "EmailSent";
-        public const string MappingAdded = "MappingAdded";
+        internal const string AccessToken = "AccessToken";
```
sam-roberts-bjss, line 5: I like the format of using the nameof feature. Could this be applied to the other fields for consistency?

tests/NHSD.BuyingCatalogue.Identity.Api.IntegrationTests/Steps/PasswordResetSteps.cs
```diff
@@ -0,0 +1,89 @@
...
+
+            userManager.RegisterTokenProvider(TokenOptions.DefaultProvider, _dataProtectionProvider);
+
+            _context[ScenarioContextKeys.IdentityUser] = identityUser;
+            _context[ScenarioContextKeys.PasswordResetToken] = await userManager.GeneratePasswordResetTokenAsync(identityUser);
+
+            await DataProtectionKeys.SaveToDb(_settings.ConnectionString, Keys);
+        }
+
+        private async Task SaveDbKeysToRepository()
```
sam-roberts-bjss, line 80: Name method as Async?

tests/NHSD.BuyingCatalogue.Identity.Api.IntegrationTests/Steps/RegistrationSteps.cs
```diff
@@ -7,19 +10,33 @@ namespace NHSD.BuyingCatalogue.Identity.Api.IntegrationTests.Steps
...
         public void WhenTheUserNavigatesToUrl(string url)
         {
-            _seleniumContext.WebDriver.Navigate().GoToUrl($"{discovery}/{url}");
+            _seleniumContext.WebDriver.Navigate().GoToUrl($"{_discovery}/{url}");
+        }
+
+        [When(@"the user navigates to identity url (\S*) with a valid password reset token")]
+        public void WhenTheUserNavigatesToUrlWithValidPasswordResetToken(string url)
+        {
+            var user = (IdentityUser)_context[ScenarioContextKeys.IdentityUser];
```
sam-roberts-bjss, line 36: Just a suggestion but the ScenarioContext has a `Get<T>` method that can retrieve a typed entry. I find it easier to read be clearly subjective, so don't feel you have to make the change.

tests/NHSD.BuyingCatalogue.Identity.Api.Testing.Data/Entities/DataProtectionKey.cs
```diff
@@ -0,0 +1,47 @@
...
+
+        public XElement Element { get; }
+
+        public string FriendlyName { get; }
+
+        public string Id { get; }
+
+        public string Xml { get; }
+
+        public bool Equals(DataProtectionKey other)
```
sam-roberts-bjss, line 29: One for the coding standards but I think it would be a good idea to use the same equality format across all objects that override the Equals method. I did it slightly different but I quite like the majority of the functionality in the `IEquatable` `Equals` method.

tests/NHSD.BuyingCatalogue.Identity.Api.Testing.Data/Entities/DataProtectionKeys.cs
```diff
@@ -0,0 +1,30 @@
...
+using System.Data.SqlClient;
+using System.Linq;
+using System.Threading.Tasks;
+using Dapper;
+
+namespace NHSD.BuyingCatalogue.Identity.Api.Testing.Data.Entities
+{
+    public static class DataProtectionKeys
+    {
+        public static async Task<IReadOnlyList<DataProtectionKey>> GetFromDb(string connectionString)
```
sam-roberts-bjss, line 11: Async & same for the method below?

tests/NHSD.BuyingCatalogue.Identity.Api.Testing.Data/Entities/DataProtectionKeys.cs
```diff
@@ -0,0 +1,30 @@
...
+        {
+            await using var databaseConnection = new SqlConnection(connectionString);
```
sam-roberts-bjss, line 13: Would it be possible to expand the `SqlRunner` class so that all our database connection logic is localised to one file? So my thinking here is to add a `QueryAsync` option unless this method can utilise the `FetchSingleResultAsync` method that already exists on the `SqlRunner` class.

src/NHSD.BuyingCatalogue.Identity.Api/Controllers/AccountController.cs
```diff
@@ -163,9 +163,11 @@ public IActionResult ForgotPasswordLinkSent()
         }
 
         [HttpGet]
-        public IActionResult ResetPassword(string email, string token)
+        public async Task<IActionResult> ResetPassword(string email, string token)
```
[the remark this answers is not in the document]
bjss-chris: I've just tried applying this change but I think we're probably better leaving it as is because the view names also need to change and I don't think it makes much sense for a view to have the `Async` suffix.

Perhaps we need to tweak our standards so that it doesn't apply to MVC controllers (obviously fine in API controllers though)...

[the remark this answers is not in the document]
sam-roberts-bjss: 👍 Saying that we have had a few issues with ending controller methods with Async. See the TrimAsync method, but yeah agreed one to review & makes sense to leave it out here.

tests/NHSD.BuyingCatalogue.Identity.Api.IntegrationTests/Steps/RegistrationSteps.cs
bjss-chris: Thanks for point that out – it's much nicer.

tests/NHSD.BuyingCatalogue.Identity.Api.Testing.Data/Entities/DataProtectionKey.cs
bjss-chris: I'll leave this as is for now then and we can discuss our preferred implementation pattern in our next standards meeting.

sam-roberts-bjss: 👍

src/NHSD.BuyingCatalogue.Identity.Api/appsettings.Development.json
```diff
@@ -50,5 +50,8 @@
   ],
   "smtpServer": {
     "allowInvalidCertificate": true
+  },
```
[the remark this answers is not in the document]
bjss-chris: I can't remember if the MS docs said that the application name was something that didn't necessarily need to be kept secret. However, it's probably better to err on the side of caution so I've added PLAT task 6640.

PR #173, opened 2020-04-29
6342 integration test http client refactor

assoomro:
[AB#6287](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/6287)

### DIFF
--- a/tests/NHSD.BuyingCatalogue.Identity.Api.IntegrationTests/Steps/Common/Request.cs
+++ b/tests/NHSD.BuyingCatalogue.Identity.Api.IntegrationTests/Steps/Common/Request.cs
@@ -1,0 +1,46 @@
+﻿using System.Threading.Tasks;
+using Flurl;
+using Flurl.Http;
+using TechTalk.SpecFlow;
+
+namespace NHSD.BuyingCatalogue.Identity.Api.IntegrationTests.Steps.Common
+{
+    public sealed class Request
+    {
+        private readonly Response _response;
+        private readonly ScenarioContext _context;
+
+        public Request(Response response, ScenarioContext context)
+        {
+            _response = response;
+            _context = context;
+        }
+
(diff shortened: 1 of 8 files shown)


tests/NHSD.BuyingCatalogue.Identity.Api.IntegrationTests/Steps/Common/Request.cs
```diff
@@ -0,0 +1,46 @@
...
+
+        public Request(Response response, ScenarioContext context)
+        {
+            _response = response;
+            _context = context;
+        }
+
+        public async Task GetAsync(string url, params object[] pathSegments)
+        {
+            _response.Result = await url
```
sam-roberts-bjss, line 21: The first 4 lines in each of these methods are duplicated. Is there a nice way of refactoring so this piece of code is not duplicated. Though I should point out it is still easy to read as it is.

assoomro: 😄 Great that you mention this! 

I've pondered that when I was writing this class and...

There is a way, however it is not as pretty.

We can have an extension method that combines those repeated steps.

It would looks something like this, however I think it takes some of the readability away.

image

If you find that its legible let me know and I can refactor it to ^ and reduce the duplication.

sam-roberts-bjss: Yeah that doesn't read that well. I was thinking more along the lines of:

image

assoomro: That looks much better. I'll go with that. Thanks!

PR #175, opened 2020-04-30
3926: Null credentials bug

RuthBerry96:


tests/NHSD.BuyingCatalogue.Identity.Api.IntegrationTests/Features/Reset Password.3926.feature
```diff
@@ -38,3 +38,9 @@ Scenario: 4. Not adhering to the password policy & entering mismatched passwords
     Then the user is redirected to page account/resetpassword
     And the element with Data ID field-reset-password has validation error with text The password you’ve entered does not meet the criteria
     And the element with Data ID field-confirm-reset-password has validation error with text Passwords do not match
+
+@3926
+Scenario: 5. Pressing submit without entering a password gives the user a relevant error message
+    When element with Data ID reset-password-button is clicked
+    Then the user is redirected to page account/resetpassword
+    And the element with Data ID field-reset-password has validation error with text Enter a password
```
bjss-chris, line 46: I think this feature file will need tweaking following the merge of the bug fix 6339 because a valid token is now required for these tests. However, a token is currently only generated once for the feature but scenario 1 will 'consume' that token so this new scenario will result in redirection to the link expired view and fail.

I think there are two possible solutions here: we swap the order of the scenarios (so scenario 1 is the last scenario) or we generate a token per scenario. The former is probably the simplest/quickest for now.

To confirm this it's probably worth merging in development and rerunning the tests before you merge this into development.

RuthBerry96: As discussed with @bjss-chris scenario 1 tests are currently ignored and all the tests pass when merged

PR #176, opened 2020-04-30
6388: Removed password on invalid Login. changed launch setting to be…

SimonLLawrence:
… in line with other projects,


src/NHSD.BuyingCatalogue.Identity.Api/Controllers/AccountController.cs
```diff
@@ -76,6 +76,7 @@ public async Task<IActionResult> Login(LoginViewModel viewModel)
 
                 if (signInErrors.Contains(LoginUserErrors.UserNameOrPasswordIncorrect()))
                 {
+                    ModelState.Remove("Password");
```
bjss-chris, line 79: Can we use `nameof(LoginViewModel.Password)` here?

PR #180, opened 2020-05-04
Change Email UI port to prevent clash with cluster

assoomro:


README.md
```diff
@@ -67,7 +67,7 @@ This server is [maildev](https://github.com/maildev/maildev) server with [stunne
 | outside the docker network | localhost                  | 1025  |  X  |
 | outside the docker network | localhost                  | 1587  | 1.2 |
 
-Navigate yourself to [localhost:1080](http://localhost:1080/) to view the mailbox UI
+Navigate yourself to [localhost:1180/email](http://localhost:1180/email) to view the mailbox UI
```
stuart-bass-bjss, line 70: I feel like it would be useful to mention inside / outside the cluster. That's probably a wider point about the whole readme though, which is looking a bit out of date.

PR #183, opened 2020-05-05
6446: Removing throw if null extensions

RuthBerry96:


src/NHSD.BuyingCatalogue.Identity.Api/Infrastructure/ValidationSummaryTagHelper.cs
```diff
@@ -65,9 +70,13 @@ private TagBuilder GetErrorListBuilder()
...
             builder.AddCssClass(TagHelperConstants.NhsList);
             builder.AddCssClass(TagHelperConstants.NhsValidationSummaryList);
-
+            
             var viewType = ViewContext.ViewData.Model.GetType();
-            viewType.ThrowIfNull();
+
+            if (viewType is null)
+            {
+                throw new InvalidOperationException();
```
sam-roberts-bjss, line 78: Previously the function would return `ArgumentNullException`, will this be OK changing the exception to an `InvalidOperationException`?

RuthBerry96: viewType isn't an passed in as an argument so the correct implementation is actually InvalidOperationException

PR #189, opened 2020-05-11
Fix specflow build fail

assoomro:
[AB#5542](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/5542)

- use latest .NET Core SDK
- update job display names to describe what they are doing more accurately
- stop building dev everytime

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,6 +1,7 @@
 variables:
   dockerVersion: '19.03.5'
   dotnetVersion: '3.x'
+  MSBUILDSINGLELOADCONTEXT: '1'
 
 trigger:
 - master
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -91,6 +92,7 @@ jobs:
 
   - task: DockerCompose@0
     displayName: 'Build: Dev'
+    condition: and(succeeded(), ne(variables['build.reason'], 'PullRequest'))
```
sam-roberts-bjss, line 95: Would you mind giving brief overview of what this change provides?

assoomro: I followed the pattern. We were only pushing to dev acr if this condition was met, therefore it seemed to me that we don't need to build it unless that condition is met.
As per the [docs](https://docs.microsoft.com/en-us/azure/devops/pipelines/process/conditions?view=azure-devops&tabs=yaml) the condition means that this step/task will only trigger if the build (previous steps ) succeeded and the reason behind this build is not a PullRequest.

assoomro:
> Looks good. Does this need to be applied to the other application pipelines e.g. Ordering.Api

I think it would make sense to do so, I've already started with Ordapi https://github.com/nhs-digital-gp-it-futures/BuyingCatalogueOrdering/pull/17

PR #192, opened 2020-05-15
changed bobs agreement value to 1

talawson05:
It would be nice if Bobs catalogue agreement was already set, so that when i run tests locally for the first time after launching, the accept agreement doesn't appear and cause the acceptance tests to fail.

Let's see if any unit/integration tests fail.

If everything passes, would there be any harm in this? If there's a valid reason for having bob's value set to 0, I'm happy to leave this out.


talawson05:
ok looks like all unit and integration tests have passed. So unless there's any reason why not, I think this should be merged

PR #201, opened 2020-05-26
6688 - Update Certificate Injection to use CRT and KEY files

stuart-bass-bjss:


src/NHSD.BuyingCatalogue.Identity.Api/Certificates/Certificate.cs
```diff
@@ -35,19 +37,56 @@ private static X509Certificate2 Load(ILogger logger, CertificateSettings setting
 
             try
             {
-                var certificate = new X509Certificate2(path, settings.CertificatePassword);
+                byte[] certBuffer = GetBytesFromPem(settings.CertificatePath, PemStringType.Certificate);
```
sam-roberts-bjss, line 40: I'm not sure the enum is needed here as you have already made the decision as to which header & footer of the certificate to use. Wouldn't it be better to create two methods to supply the appropriate header and footer to the `GetBytesFromPem` method? This way it wouldn't have to return null (switch statement) & if the `GetBytesFromPem` method requires a header and footer it is clearer. The code could also throw an argum,ent null exception if the header & footer are null as it looks like they shouldn't be.

src/NHSD.BuyingCatalogue.Identity.Api/Certificates/Certificate.cs
```diff
@@ -35,19 +37,56 @@ private static X509Certificate2 Load(ILogger logger, CertificateSettings setting
...
 
-                return certificate;
+                return certificateWithPrivateKey;
             }
             catch (Exception e)
             {
                 throw new CertificateSettingsException("Error with certificate or settings", e);
             }
         }
+        private static byte[] GetBytesFromPem(string pemString, PemStringType type)
```
sam-roberts-bjss, line 61: Spacing?

sam-roberts-bjss:
It looks like the pipeline is failing. Do we have any tests for this section of the code?

src/NHSD.BuyingCatalogue.Identity.Api/Certificates/Certificate.cs
```diff
@@ -58,35 +58,16 @@ private static X509Certificate2 Load(ILogger logger, CertificateSettings setting
                 throw new CertificateSettingsException("Error with certificate or settings", e);
             }
         }
-        private static byte[] GetBytesFromPem(string pemString, PemStringType type)
+        private static byte[] GetBytesFromPem(string pemString, string sectionToken)
```
sam-roberts-bjss, line 61: New line between methods?

sam-roberts-bjss, line 61: It might be worth adding guards to the parameters to ensure they are not null, especially `pemString` as this looks like data from external source.

stuart-bass-bjss: It's all wrapped in a try... catch block, which will handle & output a decent error message for all the different kinds of invalid input. I could add extra parameter checks, but it seems a little unnecessary

PR #212, opened 2020-06-15
7744: Updated ServiceRecipientRepository to handle empty results

CallumBradburyBJSS:
Story: [AB#4900](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/4900)
Task: [AB#7744](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/7744)

### DIFF
--- a/src/NHSD.BuyingCatalogue.Organisations.Api/Repositories/ServiceRecipientRepository.cs
+++ b/src/NHSD.BuyingCatalogue.Organisations.Api/Repositories/ServiceRecipientRepository.cs
@@ -24,35 +24,36 @@ namespace NHSD.BuyingCatalogue.Organisations.Api.Repositories
 
             var costCentres = new List<ServiceRecipient>();
             int offset = 0;
+            int searchLimit = _settings.GetChildOrganisationSearchLimit;
 
             while (!retrievedAll)
             {
-                var children = await _settings.ApiBaseUrl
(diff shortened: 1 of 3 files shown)


CallumBradburyBJSS:
> What if the ODS service returned an empty list of Organisations, would it result in an infinite loop?

It works fine if ODS returns an empty list - in fact that is the very scenario I wrote this PR for!

> The following properties are accessed in multiple places I wonder if it would simplify the code to assign them to a variable to reduce duplication?
> 
> 
> _settings.GetChildOrganisationSearchLimit
> children.Organisations

I have done this for GetChildOrganisationSearchLimit, as it makes it a bit more readable. I passed on doing it for children.Organisations though, as it starts complaining about multiple enumeration and I don't really want to do .ToList on it just to sort that out.

> Is it the responsibility of the controller to add the additional organisation & order the list by name? At first I was expecting to find this logic in the repository. Wonder if it would change if we were talking to a database.

I think it is the controllers responsibility for now, as the ServiceRecipientRepository is returning the child organisations, which the parent organisation is not a part of. It's doing a very specific thing, and then the controller is going 'ok, now I have the child organisations I can add the organisation that was sent to me'.

I think the same applies for ordering by name, only the controller cares about the order, the repository is just returning them all.

PR #217, opened 2020-06-18
7925: Updated ServiceRecipient query to only add offset if one exists

CallumBradburyBJSS:
Story: [AB#4900](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/4900)
Task: [AB#7925](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/7925)

### DIFF
--- a/src/NHSD.BuyingCatalogue.Organisations.Api/Repositories/ServiceRecipientRepository.cs
+++ b/src/NHSD.BuyingCatalogue.Organisations.Api/Repositories/ServiceRecipientRepository.cs
@@ -28,7 +28,6 @@ namespace NHSD.BuyingCatalogue.Organisations.Api.Repositories
 
             while (!retrievedAll)
             {
-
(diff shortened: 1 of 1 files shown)


src/NHSD.BuyingCatalogue.Organisations.Api/Repositories/ServiceRecipientRepository.cs
```diff
@@ -28,15 +28,21 @@ public async Task<IEnumerable<ServiceRecipient>> GetServiceRecipientsByParentOds
 
             while (!retrievedAll)
             {
-                var serviceRecipientResponse = await _settings.ApiBaseUrl
+
+                var query = _settings.ApiBaseUrl
```
RuthBerry96, line 32: Can you sort out the spacing here please

PR #227, opened 2020-08-18
9069: Update footer

RuthBerry96:
Story: [AB#9069](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/9069)
Task: [AB#9483](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/9483)


sam-roberts-bjss (changes requested):
Looks good 👍 We should already have some UI tests to make sure the footer content is rendered as expected. Will need to add a test for the Terms & conditions.

sam-roberts-bjss (approved):
👍

PR #228, opened 2020-08-19
9069: Update footer PR Comment

RuthBerry96:
Story: [AB#9069](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/9069)
Task: [AB#9483](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/9483)


sam-roberts-bjss (approved):
We really need to review those footer tests. Shouldn't have to add / change that test in so many files. I think it would be better to have a set of tests for the footer to check the content. Then have the individual page tests just ensure the footer is rendered & not check the content. Would you mind raising a tech debt ticket for this? It can go on the backlog

PR #252, opened 2021-02-17
Add StyleCop to Identity API project

bjss-chris:
Activity: [AB#9815](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/9815)
Task: [AB#11099](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/11099)


src/NHSD.BuyingCatalogue.Identity.Api/Controllers/AccountController.cs
```diff
@@ -35,24 +35,21 @@ public sealed class AccountController : Controller
...
         {
-            if (returnUrl == null)
+            if (returnUrl is null)
                 return RedirectToPublicBrowseLogin();
 
-            LoginViewModel loginViewModel = new LoginViewModel
-            {
-                ReturnUrl = returnUrl
-            };
+            LoginViewModel loginViewModel = new LoginViewModel { ReturnUrl = returnUrl };
```
PhysicsJon, line 52: can this be updated to `new()`?

PhysicsJon (changes requested):
Just a single little change

src/NHSD.BuyingCatalogue.Identity.Api/Controllers/AccountController.cs
bjss-chris: Yes, it can although I didn't get any prompts for it (either from the built-in VS analyser or ReSharper) for some reason, which is why I missed it. However, rather than use `new()` I'll use `var` instead.

bjss-chris: Good spot by the way!

PhysicsJon (approved):

PR #302, opened 2021-04-12
Feature 12947 define proxy organisation relationships

leerogers-bjss:
Added Get RelatedOrganisation, Get UnrelatedOrganisation and Create New RelatedOrganisation Endpoints

Story : [AB#12947](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/12947)

### DIFF
--- a/src/NHSD.BuyingCatalogue.Organisations.Api/Controllers/OrganisationsController.cs
+++ b/src/NHSD.BuyingCatalogue.Organisations.Api/Controllers/OrganisationsController.cs
@@ -247,12 +247,12 @@ namespace NHSD.BuyingCatalogue.Organisations.Api.Controllers
 
             if (relatedOrganisation is null)
             {
-                return BadRequest(new ErrorMessageViewModel(FormattableString.Invariant($"The referenced Organisation {model.RelatedOrganisationId} cannot be found.")));
+                return BadRequest(new ErrorMessageViewModel(FormattableString.Invariant($"The referenced organisation {model.RelatedOrganisationId} cannot be found.")));
             }
 
             if (organisation.RelatedOrganisations.Any(ro => ro.OrganisationId == model.RelatedOrganisationId))
             {
-                return BadRequest(new ErrorMessageViewModel(FormattableString.Invariant($"The referenced Organisation {model.RelatedOrganisationId} is already Related to {organisation.OrganisationId}.")));
+                return BadRequest(new ErrorMessageViewModel(FormattableString.Invariant($"The referenced organisation {model.RelatedOrganisationId} is already related to {organisation.OrganisationId}.")));
             }
 
             organisation.RelatedOrganisations.Add(relatedOrganisation);
--- a/src/NHSD.BuyingCatalogue.Organisations.Api/Controllers/OrganisationsController.cs
+++ b/src/NHSD.BuyingCatalogue.Organisations.Api/Controllers/OrganisationsController.cs
@@ -230,8 +230,7 @@ namespace NHSD.BuyingCatalogue.Organisations.Api.Controllers
                 var listOfRelatedOrganisationsGuids = organisation.RelatedOrganisations
                                                                   .Select(ro => ro.ChildOrganisation).Select(ro => ro.OrganisationId).ToList();
 
-                var unrelatedOrganisations = organisationsList.Where(o => !listOfRelatedOrganisationsGuids
-                                                              .Any(ro => ro == o.OrganisationId))
+                var unrelatedOrganisations = organisationsList.Where(o => !listOfRelatedOrganisationsGuids.Any(ro => ro == o.OrganisationId))
                                                               .Where(o => o.OrganisationId != organisation.OrganisationId); // second where to clear out calling org
 
                 return unrelatedOrganisations.Select(uo => new RelatedOrganisationModel() { OrganisationId = uo.OrganisationId, Name = uo.Name, OdsCode = uo.OdsCode }).ToList();
--- a/src/NHSD.BuyingCatalogue.Organisations.Api/Controllers/OrganisationsController.cs
+++ b/src/NHSD.BuyingCatalogue.Organisations.Api/Controllers/OrganisationsController.cs
@@ -207,7 +207,9 @@ namespace NHSD.BuyingCatalogue.Organisations.Api.Controllers
                 return NotFound();
             }
 
-            return organisation.RelatedOrganisations.Select(ro => ro.ChildOrganisation).Select(ro => new RelatedOrganisationModel() { OrganisationId = ro.OrganisationId, Name = ro.Name, OdsCode = ro.OdsCode }).ToList();
+            return organisation.RelatedOrganisations.Select(ro => ro.ChildOrganisation)
+                                                    .Select(ro => new RelatedOrganisationModel() { OrganisationId = ro.OrganisationId, Name = ro.Name, OdsCode = ro.OdsCode })
+                                                    .ToList();
         }
 
         [HttpGet]
@@ -225,9 +227,12 @@ namespace NHSD.BuyingCatalogue.Organisations.Api.Controllers
 
             if (organisation.RelatedOrganisations is not null || organisation.RelatedOrganisations.Count > 0)
             {
-                var listOfRelatedOrganisationsGuids = organisation.RelatedOrganisations.Select(ro => ro.ChildOrganisation).Select(ro => ro.OrganisationId).ToList();
+                var listOfRelatedOrganisationsGuids = organisation.RelatedOrganisations
+                                                                  .Select(ro => ro.ChildOrganisation).Select(ro => ro.OrganisationId).ToList();
 
-                var unrelatedOrganisations = organisationsList.Where(o => !listOfRelatedOrganisationsGuids.Any(ro => ro == o.OrganisationId)).Where(o => o.OrganisationId != organisation.OrganisationId); // second where to clear out calling org
+                var unrelatedOrganisations = organisationsList.Where(o => !listOfRelatedOrganisationsGuids
+                                                              .Any(ro => ro == o.OrganisationId))
+                                                              .Where(o => o.OrganisationId != organisation.OrganisationId); // second where to clear out calling org
 
                 return unrelatedOrganisations.Select(uo => new RelatedOrganisationModel() { OrganisationId = uo.OrganisationId, Name = uo.Name, OdsCode = uo.OdsCode }).ToList();
             }
--- a/src/NHSD.BuyingCatalogue.Organisations.Api/Controllers/OrganisationsController.cs
+++ b/src/NHSD.BuyingCatalogue.Organisations.Api/Controllers/OrganisationsController.cs
@@ -202,7 +202,7 @@ namespace NHSD.BuyingCatalogue.Organisations.Api.Controllers
         {
             var organisation = await organisationRepository.GetByIdWithRelatedOrganisationsAsync(id);
 
-            if (organisation is null || organisation.RelatedOrganisations is null)
+            if (organisation is null || organisation.RelatedOrganisations is null || organisation.RelatedOrganisations.Count == 0)
             {
                 return NotFound();
             }
@@ -223,7 +223,7 @@ namespace NHSD.BuyingCatalogue.Organisations.Api.Controllers
 
             IEnumerable<Organisation> organisationsList = await organisationRepository.ListOrganisationsAsync();
 
-            if (organisation.RelatedOrganisations is not null)
+            if (organisation.RelatedOrganisations is not null || organisation.RelatedOrganisations.Count > 0)
             {
                 var listOfRelatedOrganisationsGuids = organisation.RelatedOrganisations.Select(ro => ro.ChildOrganisation).Select(ro => ro.OrganisationId).ToList();
 
@@ -245,7 +245,7 @@ namespace NHSD.BuyingCatalogue.Organisations.Api.Controllers
             if (model is null)
                 throw new ArgumentNullException(nameof(model));
 
-            var organisation = await organisationRepository.GetByIdWithRelatedOrganisationsAsync(id, false);
+            var organisation = await organisationRepository.GetByIdWithRelatedOrganisationsAsync(id);
 
             if (organisation is null)
             {
@@ -264,7 +264,7 @@ namespace NHSD.BuyingCatalogue.Organisations.Api.Controllers
                 return BadRequest();
             }
 
-            var newRelatedOrganisation = new RelatedOrganisation() { Organisation = organisation, ChildOrganisation = relatedOrganisation};
+            var newRelatedOrganisation = new RelatedOrganisation() { Organisation = organisation, ChildOrganisation = relatedOrganisation };
 
             organisation.RelatedOrganisations.Add(newRelatedOrganisation);
 
--- a/src/NHSD.BuyingCatalogue.Organisations.Api/Controllers/OrganisationsController.cs
+++ b/src/NHSD.BuyingCatalogue.Organisations.Api/Controllers/OrganisationsController.cs
@@ -202,14 +202,14 @@ namespace NHSD.BuyingCatalogue.Organisations.Api.Controllers
         {
             var organisation = await organisationRepository.GetByIdWithRelatedOrganisationsAsync(id);
 
-            if (organisation is null || organisation.RelatedOrganisations is null || organisation.RelatedOrganisations.Count == 0)
+            if (organisation is null)
             {
                 return NotFound();
             }
 
-            return organisation.RelatedOrganisations.Select(ro => ro.ChildOrganisation)
-                                                    .Select(ro => new RelatedOrganisationModel() { OrganisationId = ro.OrganisationId, Name = ro.Name, OdsCode = ro.OdsCode })
-                                                    .ToList();
+            return organisation.RelatedOrganisations
+                .Select(ro => new RelatedOrganisationModel { OrganisationId = ro.OrganisationId, Name = ro.Name, OdsCode = ro.OdsCode })
+                .ToList();
         }
 
         [HttpGet]
@@ -223,22 +223,9 @@ namespace NHSD.BuyingCatalogue.Organisations.Api.Controllers
                 return NotFound();
             }
 
-            IEnumerable<Organisation> organisationsList = await organisationRepository.ListOrganisationsAsync();
+            var unrelatedOrganisations = await organisationRepository.GetUnrelatedOrganisations(organisation);
 
-            if (organisation.RelatedOrganisations is not null || organisation.RelatedOrganisations.Count > 0)
-            {
-                var listOfRelatedOrganisationsGuids = organisation.RelatedOrganisations
-                                                                  .Select(ro => ro.ChildOrganisation).Select(ro => ro.OrganisationId).ToList();
-
-                var unrelatedOrganisations = organisationsList.Where(o => !listOfRelatedOrganisationsGuids.Any(ro => ro == o.OrganisationId))
-                                                              .Where(o => o.OrganisationId != organisation.OrganisationId); // second where to clear out calling org
-
-                return unrelatedOrganisations.Select(uo => new RelatedOrganisationModel() { OrganisationId = uo.OrganisationId, Name = uo.Name, OdsCode = uo.OdsCode }).ToList();
-            }
-
-            return organisationsList.Where(o => o.OrganisationId != organisation.OrganisationId)
-                                    .Select(uo => new RelatedOrganisationModel() { OrganisationId = uo.OrganisationId, Name = uo.Name, OdsCode = uo.OdsCode })
-                                    .ToList();
+            return unrelatedOrganisations.Select(uo => new RelatedOrganisationModel() { OrganisationId = uo.OrganisationId, Name = uo.Name, OdsCode = uo.OdsCode }).ToList();
         }
 
         [Authorize(Policy = PolicyName.CanManageOrganisations)]
@@ -260,17 +247,15 @@ namespace NHSD.BuyingCatalogue.Organisations.Api.Controllers
 
             if (relatedOrganisation is null)
             {
-                return BadRequest();
+                return BadRequest(new ErrorMessageViewModel(FormattableString.Invariant($"The referenced Organisation {model.RelatedOrganisationId} cannot be found.")));
             }
 
-            if (organisation.RelatedOrganisations.Any(ro => ro.RelatedOrganisationId == model.RelatedOrganisationId))
+            if (organisation.RelatedOrganisations.Any(ro => ro.OrganisationId == model.RelatedOrganisationId))
             {
-                return BadRequest();
+                return BadRequest(new ErrorMessageViewModel(FormattableString.Invariant($"The referenced Organisation {model.RelatedOrganisationId} is already Related to {organisation.OrganisationId}.")));
             }
 
-            var newRelatedOrganisation = new RelatedOrganisation() { Organisation = organisation, ChildOrganisation = relatedOrganisation };
-
-            organisation.RelatedOrganisations.Add(newRelatedOrganisation);
+            organisation.RelatedOrganisations.Add(relatedOrganisation);
 
             await organisationRepository.UpdateAsync(organisation);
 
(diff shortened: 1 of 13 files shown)


src/NHSD.BuyingCatalogue.Organisations.Api/Repositories/OrganisationRepository.cs
```diff
@@ -27,6 +27,11 @@ public async Task<Organisation> GetByIdAsync(Guid id)
             return await context.Organisations.FirstOrDefaultAsync(org => org.OrganisationId == id);
         }
 
+        public async Task<Organisation> GetByIdWithRelatedOrganisationsAsync(Guid id)
+        {
+            return await context.Organisations.Include(o => o.RelatedOrganisations).ThenInclude(o => o.ChildOrganisation).FirstOrDefaultAsync(org => org.OrganisationId == id);
+        }
+
```
bjss-chris, line 34: To simplify the `OrganisationController.GetUnrelatedOrganisationsAsync` method I'm wondering if it might be worth introducing an additional method, something like `Task<IEnumerable<Organisation>> GetUnrelatedOrganisationsAsync(Guid id)`?

I'm just thinking it might make the controller method simpler if we offloaded most of the work to DB, as if we were writing a query similar to this:

```sql
SELECT * -- Just for illustration!
  FROM dbo.Organisations AS o
 WHERE o.OrganisationId <> @id
   AND o.OrganisationId NOT IN
       (SELECT r.RelatedOrganisationId FROM dbo.RelatedOrganisations AS r WHERE r.OrganisationId = @id);
```

If we add a `RelatedOrganisations` property to the `ApplicationDbContext` I think we can probably do something like this:

```c#
return await context.Organisations
	.Where(o => o.OrganisationId != id)
	.Where(o => !context.RelatedOrganisations.Where(r => r.OrganisationId == id).Select(r => r.RelatedOrganisationId).Contains(o.OrganisationId))
	.ToListAsync();
```

I've not spent a huge amount of time thinking about it so there may be a more elegant way of doing it.

We'd probably want to switch to using `GetByIdAsync` for the first part of the controller method to check the organisation exists.

Admittedly we will have 2 DB calls with this approach but I don't think it's a problem for how frequently this is going to be invoked.

I guess there may even be a way to get it down to 1 DB call but I'll leave that for you to think about (as I don't have time)...

What do you think? Hopefully you'll be able to translate my waffle! 😃

src/NHSD.BuyingCatalogue.Organisations.Api/ViewModels/Organisations/CreateRelatedOrganisationModel.cs
```diff
@@ -0,0 +1,9 @@
+﻿using System;
+
+namespace NHSD.BuyingCatalogue.Organisations.Api.ViewModels.Organisations
+{
+    public class CreateRelatedOrganisationModel
```
bjss-chris, line 5: `sealed`.

Can we use `init` rather than `set` here?

src/NHSD.BuyingCatalogue.Organisations.Api/ViewModels/Organisations/RelatedOrganisationModel.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace NHSD.BuyingCatalogue.Organisations.Api.ViewModels.Organisations
+{
+    public class RelatedOrganisationModel
```
bjss-chris, line 8: `sealed`.

Can we use `init` rather than `set` here?

tests/NHSD.BuyingCatalogue.Organisations.Api.UnitTests/Builders/OrganisationBuilder.cs
```diff
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Collections.Generic;
```
bjss-chris, line 2: Do we need this `using` statement?

tests/NHSD.BuyingCatalogue.Organisations.Api.UnitTests/Builders/OrganisationBuilder.cs
```diff
@@ -66,9 +68,21 @@ internal OrganisationBuilder WithAddress(Address value)
             return this;
         }
 
+        internal OrganisationBuilder WithRelatedOrganisation(Guid relatedOrganisationId)
+        {
+            relatedOrganisation = new()
```
bjss-chris, line 73: I think it might be slightly clearer to include the type here (but think `new()` is fine elsewhere, e.g. if we were doing something like `return new()` here).

tests/NHSD.BuyingCatalogue.Organisations.Api.UnitTests/Builders/OrganisationBuilder.cs
```diff
@@ -78,6 +92,28 @@ internal Organisation Build()
                 Address = address,
                 LastUpdated = lastUpdated,
             };
+
+            if (relatedOrganisation != null)
```
bjss-chris, line 96: Code style is to prefer `is not null` to `!= null`.

Probably ought to document some of this stuff and it might seem a tad neurotic again! However, as before keeps things consistent.

tests/NHSD.BuyingCatalogue.Organisations.Api.UnitTests/Builders/OrganisationBuilder.cs
```diff
@@ -78,6 +92,28 @@ internal Organisation Build()
...
+            {
+                org.RelatedOrganisations.Add(relatedOrganisation);
+            }
+
+            return org;
+        }
+
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("StyleCop.CSharp.OrderingRules", "SA1204:Static elements should appear before instance elements", Justification = "This Doesn't Actually need to be at the top")]
+        internal static Organisation BuildRelatedOrganisation(Guid index)
```
bjss-chris, line 105: This can be `private`, so the suppression can be removed.

Also, occasionally we might suppress an FxCop warning but we generally don't suppress style warnings.

tests/NHSD.BuyingCatalogue.Organisations.Api.UnitTests/Controllers/OrganisationControllerTests.cs
```diff
@@ -410,5 +410,231 @@ public static async Task GetServiceRecipientsAsync_OrganisationExists_ReturnsThe
...
+
+            var result = await controller.GetRelatedOrganisationsAsync(organisation.OrganisationId);
+
+            result.Should().BeEquivalentTo(new ActionResult<RelatedOrganisationModel>(new NotFoundResult()));
+        }
+
+        [Test]
+        public static async Task GetRelatedOrganisationsAsync_OrganisationDoesNotExist_ReturnsNotFound()
+        {
+            Guid nonExistantOrganisationId = Guid.NewGuid();
```
bjss-chris, line 459: Typo, should be `Existent`.

tests/NHSD.BuyingCatalogue.Organisations.Api.UnitTests/Controllers/OrganisationControllerTests.cs
```diff
@@ -410,5 +410,231 @@ public static async Task GetServiceRecipientsAsync_OrganisationExists_ReturnsThe
 
             response.Should().BeEquivalentTo(new ActionResult<List<ServiceRecipientsModel>>(expected), config => config.WithStrictOrdering());
         }
+
+        [Test]
+        public static async Task GetOrganisationsAsync_OrganisationExists_ReturnsOrganisationAndRelatedOrganisation()
+        {
+            Guid relatedOrganisationId = Guid.NewGuid();
```
bjss-chris, line 417: In contrast to my earlier `new()` comment, I think `var` would be fine here. It's fairly obviously that it's going to be a `Guid`. 😃

(Same applies to the similar line in other tests.)

tests/NHSD.BuyingCatalogue.Organisations.Api.UnitTests/Controllers/OrganisationControllerTests.cs
```diff
@@ -410,5 +410,231 @@ public static async Task GetServiceRecipientsAsync_OrganisationExists_ReturnsThe
...
+
+            var expected = new List<RelatedOrganisationModel>();
+
+            result.Should().BeEquivalentTo(new ActionResult<List<RelatedOrganisationModel>>(expected));
+        }
+
+        [Test]
+        public static async Task GetUnrelatedOrganisationAsync_OrganisationDoesNotExist_ReturnsNotFound()
+        {
+            Guid nonExistantOrganisationId = Guid.NewGuid();
```
bjss-chris, line 560: Typo, as above.

tests/NHSD.BuyingCatalogue.Organisations.Api.UnitTests/Controllers/OrganisationControllerTests.cs
```diff
@@ -410,5 +410,231 @@ public static async Task GetServiceRecipientsAsync_OrganisationExists_ReturnsThe
...
+            var organisation = OrganisationBuilder.Create(1).Build();
+
+            Guid nonExistingRelatedOrganisationId = Guid.NewGuid();
+
+            var controller = OrganisationControllerBuilder
+                .Create()
+                .WithGetOrganisationWithRelatedOrganisations(organisation)
+                .Build();
+
+            var result = await controller.CreateRelatedOrganisationAsync(organisation.OrganisationId, new CreateRelatedOrganisationModel() { RelatedOrganisationId = nonExistingRelatedOrganisationId });
```
bjss-chris, line 616: `()` can be omitted from `new CreateRelatedOrganisationModel()`.

tests/NHSD.BuyingCatalogue.Organisations.Api.UnitTests/Controllers/OrganisationControllerTests.cs
```diff
@@ -410,5 +410,231 @@ public static async Task GetServiceRecipientsAsync_OrganisationExists_ReturnsThe
...
+            var organisation = OrganisationBuilder.Create(1).WithRelatedOrganisation(relatedOrganisationId).Build();
+
+            var relatedOrganisation = OrganisationBuilder.Create(2).WithOrganisationId(relatedOrganisationId).Build();
+
+            var controller = OrganisationControllerBuilder
+                .Create()
+                .WithGetByIdWithRelatedAndGetByIdForRelatedAndUpdateAsync(organisation, relatedOrganisation)
+                .Build();
+
+            var result = await controller.CreateRelatedOrganisationAsync(organisation.OrganisationId, new CreateRelatedOrganisationModel() { RelatedOrganisationId = relatedOrganisationId });
```
bjss-chris, line 635: `()` can be omitted from `new CreateRelatedOrganisationModel()`.

bjss-chris (changes requested):
I've left a few comments. Most of them are fairly small things with one slightly more substantial suggestion/question – let me know if any questions.

src/NHSD.BuyingCatalogue.Organisations.Api/Controllers/OrganisationsController.cs
```diff
@@ -195,5 +195,86 @@ public async Task<ActionResult<IEnumerable<ServiceRecipientsModel>>> GetServiceR
...
+        [HttpGet]
+        [Route("{id}/related-organisations")]
+        public async Task<ActionResult<IEnumerable<RelatedOrganisationModel>>> GetRelatedOrganisationsAsync(Guid id)
+        {
+            var organisation = await organisationRepository.GetByIdWithRelatedOrganisationsAsync(id);
+
+            if (organisation is null || organisation.RelatedOrganisations is null || organisation.RelatedOrganisations.Count == 0)
+            {
+                return NotFound();
+            }
```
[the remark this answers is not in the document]
leerogers-bjss: yeah, this was a check before we were substantiating the list it was nullable.
i kept it in for sanities sake, but will remove.

src/NHSD.BuyingCatalogue.Organisations.Api/Data/ApplicationDbContext.cs
```diff
@@ -18,6 +18,16 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
...
+                         .HasOne(or => or.ChildOrganisation)
+                         .WithMany()
+                         .HasForeignKey(or => or.RelatedOrganisationId);
+
+            modelBuilder.Entity<RelatedOrganisation>()
+                        .HasOne(or => or.Organisation)
+                        .WithMany(o => o.RelatedOrganisations)
+                        .HasForeignKey(or => or.OrganisationId);
+
             base.OnModelCreating(modelBuilder);
```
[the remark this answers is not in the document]
leerogers-bjss: i was trying to keep away from setting up a DbSet for RelatedOrgs

in reality, i would have loved this table and it's object to never really be seen at all and just get a pure list of orgs, but that's not how EF likes to do things.  

this code seems to be the EF way of doing code first linked tables. EF core can automagically work it out from Table Names, but then you have to name your silly stuff, so instead of RelatedOrganisation table, we would have OrganisationOrganisation table which would be super confusing. this is kind of the best way for EF to automatically pick it up without doing crazy stuff.

then again, i could hand roll it, but you get the advantage this way of being able to just add and remove the RelatedOrganisation from the list inside of an Organisation and it handles the adding/deletion pretty simply.

tbh, i don't mind either way, this just seemed to be the more EF way of things.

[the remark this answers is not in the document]
bjss-chris: We hand-roll the entity set-up in the ordering API so feel free to consider that option.

src/NHSD.BuyingCatalogue.Organisations.Api/Repositories/OrganisationRepository.cs
leerogers-bjss: i can translate :D i did think about doing this, but theres examples of this work being done in the controller, so took that as the leader. however, i think this function should suffice.

`        public async Task<IEnumerable<Organisation>> GetUnrelatedOrganisations(Organisation organisation)
        {
            return await context.Organisations
                .Where(o => o.OrganisationId != organisation.OrganisationId)
                .Where(o => !organisation.RelatedOrganisations.Any(ro => o.OrganisationId == ro.RelatedOrganisationId))
                .OrderBy(o => o.Name)
                .ToListAsync();
        }`
        
        technically we can't get around the 2 DB calls. if we did a call that gets unrelated on an non-existent org, we would get back a blank list, which would be the same if the org was related to every other org.
        
        so the first call should be to check if the org you're calling exists, then we can call the function above on that org to  get unrelated.
        
        seem ok to you?

bjss-chris: Yep, seems OK to me! 😃

tests/NHSD.BuyingCatalogue.Organisations.Api.UnitTests/Controllers/OrganisationControllerTests.cs
```diff
@@ -410,5 +410,134 @@ public static async Task GetServiceRecipientsAsync_OrganisationExists_ReturnsThe
...
+        }
+
+        [Test]
+        public static async Task CreateRelatedOrganisationAsync_RelatedOrganisationReturnsSuccess_ReturnsRelatedOrganisation()
+        {
+            var organisation = OrganisationBuilder.Create(1).Build();
+
+            var relatedOrganisation = OrganisationBuilder.Create(2).Build();
+
+            var relatedOrganisationModel = new CreateRelatedOrganisationModel() { RelatedOrganisationId = relatedOrganisation.OrganisationId };
```
bjss-chris, line 477: `()` can be omitted from `new CreateRelatedOrganisationModel()`.

tests/NHSD.BuyingCatalogue.Organisations.Api.UnitTests/Controllers/OrganisationControllerTests.cs
```diff
@@ -144,9 +144,9 @@ public static async Task GetAllAsync_ListOfOrganisationsExist_ReturnsTheOrganisa
 
             var organisationList = organisationResult.Organisations.ToList();
 
-            organisationList[0].Should().BeEquivalentTo(org1, config => config.Excluding(o => o.LastUpdated));
-            organisationList[1].Should().BeEquivalentTo(org2, config => config.Excluding(o => o.LastUpdated));
-            organisationList[2].Should().BeEquivalentTo(org3, config => config.Excluding(o => o.LastUpdated));
+            organisationList[0].Should().BeEquivalentTo(org1, config => config.Excluding(o => o.LastUpdated).Excluding(o => o.RelatedOrganisations).Excluding(o => o.ParentRelatedOrganisations));
+            organisationList[1].Should().BeEquivalentTo(org2, config => config.Excluding(o => o.LastUpdated).Excluding(o => o.RelatedOrganisations).Excluding(o => o.ParentRelatedOrganisations));
+            organisationList[2].Should().BeEquivalentTo(org3, config => config.Excluding(o => o.LastUpdated).Excluding(o => o.RelatedOrganisations).Excluding(o => o.ParentRelatedOrganisations));
```
bjss-chris, line 149: We don't currently have any formal style rules around line length but I think these are probably getting a bit on the long side. Recommend breaking like this:

```c#
organisationList[0].Should().BeEquivalentTo(
    org1,
    config => config.Excluding(o => o.LastUpdated).Excluding(o => o.RelatedOrganisations).Excluding(o => o.ParentRelatedOrganisations));

organisationList[1].Should().BeEquivalentTo(// ...
```

tests/NHSD.BuyingCatalogue.Organisations.Api.UnitTests/Controllers/OrganisationControllerTests.cs
```diff
@@ -410,5 +410,134 @@ public static async Task GetServiceRecipientsAsync_OrganisationExists_ReturnsThe
...
+                .Create()
+                .Build();
+
+            var result = await controller.CreateRelatedOrganisationAsync(organisationId, new CreateRelatedOrganisationModel());
+
+            result.Should().BeOfType<NotFoundResult>();
+        }
+
+        [Test]
+        public static async Task CreateRelatedOrganisationsAsync_OrganisationExists_RelatedOrganisationDoesntExist_ReturnsBadRequest()
```
bjss-chris, line 504: Think it would be better if we spell out `DoesNot` here rather than use the contracted form.

bjss-chris (changes requested):
Mostly small changes suggested but one question around EF Core support for the unrelated organisations implementation.

src/NHSD.BuyingCatalogue.Organisations.Api/Data/ApplicationDbContext.cs
```diff
@@ -18,6 +18,21 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
...
+            modelBuilder.Entity<Organisation>()
+                .HasMany(o => o.RelatedOrganisations)
+                .WithMany(o => o.ParentRelatedOrganisations)
+                .UsingEntity<RelatedOrganisation>(
+                ro => ro.HasOne(ro => ro.Organisation)
+                      .WithMany()
+                      .HasForeignKey(ro => ro.OrganisationId)
+                      .HasConstraintName("FK_RelatedOrganisations_OrganisationId")
+                      .OnDelete(DeleteBehavior.Cascade),
+                ro => ro.HasOne(ro => ro.ChildOrganisation)
```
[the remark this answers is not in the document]
leerogers-bjss: could always go back to x 😆

src/NHSD.BuyingCatalogue.Organisations.Api/Repositories/OrganisationRepository.cs
```diff
@@ -27,6 +27,20 @@ public async Task<Organisation> GetByIdAsync(Guid id)
...
+        public async Task<Organisation> GetByIdWithRelatedOrganisationsAsync(Guid id)
+        {
+            return await context.Organisations.Include(o => o.RelatedOrganisations).FirstOrDefaultAsync(org => org.OrganisationId == id);
+        }
+
+        public async Task<IEnumerable<Organisation>> GetUnrelatedOrganisations(Organisation organisation)
+        {
+            return await context.Organisations
+                .Where(o => o.OrganisationId != organisation.OrganisationId)
+                .Where(o => !organisation.RelatedOrganisations.Any(ro => o.OrganisationId == ro.OrganisationId))
```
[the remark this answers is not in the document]
leerogers-bjss: yeah, its generating SQL.
Which is why i can't test it. see the slack message i sent you this morning

tests/NHSD.BuyingCatalogue.Organisations.Api.UnitTests/Controllers/OrganisationControllerTests.cs
leerogers-bjss: well theres a couple of ways to skin this cat, i would suggest this:

```
            organisationList[0].Should().BeEquivalentTo(
                org1,
                config => config.Excluding(o => o.LastUpdated)
                                .Excluding(o => o.RelatedOrganisations)
                                .Excluding(o => o.ParentRelatedOrganisations));
```
thoughts?

src/NHSD.BuyingCatalogue.Organisations.Api/Repositories/OrganisationRepository.cs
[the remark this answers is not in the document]
leerogers-bjss: actually, this fails if you don't have any related organisations. just doing a fixed version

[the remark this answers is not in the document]
leerogers-bjss: i've just updated the GetUnrelatedOrganisations to this:

```c#
        public async Task<IEnumerable<Organisation>> GetUnrelatedOrganisations(Organisation organisation)
        {
            return await context.Organisations.Where(o => o != organisation && !o.ParentRelatedOrganisations.Contains(organisation)).ToListAsync();
        }
```
which is generating this SQL
```sql
DECLARE @__entity_equality_organisation_0_OrganisationId uniqueIdentifier = '9173b946-793a-4b28-906d-270db18dd82a';

SELECT [o].[OrganisationId], [o].[Address], [o].[CatalogueAgreementSigned], [o].[LastUpdated], [o].[Name], [o].[OdsCode], [o].[PrimaryRoleId]
FROM [Organisations] AS [o]
WHERE ([o].[OrganisationId] <> @__entity_equality_organisation_0_OrganisationId) AND NOT (EXISTS (
    SELECT 1
    FROM [RelatedOrganisations] AS [r]
    LEFT JOIN [Organisations] AS [o0] ON [r].[OrganisationId] = [o0].[OrganisationId]
    WHERE ([o].[OrganisationId] = [r].[RelatedOrganisationId]) AND ([o0].[OrganisationId] = @__entity_equality_organisation_0_OrganisationId)))
    ```
    
    which seems fine to me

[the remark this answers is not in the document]
Comment (bjss-chris) [reply src/NHSD.BuyingCatalogue.Organisations.Api/Repositories/OrganisationRepository.cs:39]: Yeah, that looks good to me.

File: tests/NHSD.BuyingCatalogue.Organisations.Api.UnitTests/Controllers/OrganisationControllerTests.cs (hunk 16 above)
Comment (bjss-chris) [reply tests/NHSD.BuyingCatalogue.Organisations.Api.UnitTests/Controllers/OrganisationControllerTests.cs:149]: Your suggestion is alright but would prefer it with a slight tweak:

```c#
organisationList[0].Should().BeEquivalentTo(
    org1,
    config => config
        .Excluding(o => o.LastUpdated)
        .Excluding(o => o.RelatedOrganisations)
        .Excluding(o => o.ParentRelatedOrganisations));
```

I prefer this because it doesn't cause my eyes to scan as far to the right. A small, and subjective, thing but I find that a bit more readable – it seems more natural to me.

bjss-chris (approved):
We need to add some integration tests but let's create a separate task for that.

Can you also squash your commits? Probably down to just one should be fine, I think.

PR #305, opened 2021-04-15
Added DeleteRelatedOrganisationAsync endpoint.

leerogers-bjss:
updated all route references from {id} to {organisationId} in organisationsController to make the swagger/referenced object clearer.
added tests for DeleteRelatedOrganisationAsync endpoint

Story:  [AB#12954](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/12954)
Task: [AB#13820](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/13820)

### DIFF
--- a/src/NHSD.BuyingCatalogue.Organisations.Api/Controllers/OrganisationsController.cs
+++ b/src/NHSD.BuyingCatalogue.Organisations.Api/Controllers/OrganisationsController.cs
@@ -67,10 +67,10 @@ namespace NHSD.BuyingCatalogue.Organisations.Api.Controllers
         }
 
         [HttpGet]
-        [Route("{organisationId}")]
-        public async Task<ActionResult> GetByIdAsync(Guid organisationId)
+        [Route("{id}")]
+        public async Task<ActionResult> GetByIdAsync(Guid id)
         {
-            var organisation = await organisationRepository.GetByIdAsync(organisationId);
+            var organisation = await organisationRepository.GetByIdAsync(id);
 
             if (organisation is null)
             {
@@ -102,15 +102,15 @@ namespace NHSD.BuyingCatalogue.Organisations.Api.Controllers
 
         [Authorize(Policy = PolicyName.CanManageOrganisations)]
         [HttpPut]
-        [Route("{organisationId}")]
-        public async Task<ActionResult> UpdateOrganisationByIdAsync(Guid organisationId, UpdateOrganisationModel model)
+        [Route("{id}")]
+        public async Task<ActionResult> UpdateOrganisationByIdAsync(Guid id, UpdateOrganisationModel model)
         {
             if (model is null)
             {
                 throw new ArgumentNullException(nameof(model));
             }
 
-            var organisation = await organisationRepository.GetByIdAsync(organisationId);
+            var organisation = await organisationRepository.GetByIdAsync(id);
 
             if (organisation is null)
             {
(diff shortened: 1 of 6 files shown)


src/NHSD.BuyingCatalogue.Organisations.Api/Controllers/OrganisationsController.cs
```diff
@@ -261,5 +261,31 @@ public async Task<ActionResult> CreateRelatedOrganisationAsync(Guid id, [FromBod
...
+
+            if (organisation is null)
+            {
+                return NotFound();
+            }
+
+            if (!organisation.RelatedOrganisations.Any(ro => ro.OrganisationId == relatedOrganisationId))
+            {
+                return BadRequest(new ErrorMessageViewModel(FormattableString.Invariant($"The referenced organisation {organisationId} has no relationship to {relatedOrganisationId}.")));
+            }
```
bjss-chris, line 280: I think we can omit this. I think a `NoContent()` will be fine in this case – please see additional comments below.

src/NHSD.BuyingCatalogue.Organisations.Api/Controllers/OrganisationsController.cs
```diff
@@ -261,5 +261,31 @@ public async Task<ActionResult> CreateRelatedOrganisationAsync(Guid id, [FromBod
...
+
+            var relatedOrganisation = organisation.RelatedOrganisations.Where(ro => ro.OrganisationId == relatedOrganisationId).First();
```
bjss-chris, line 282: Semantically, I think `Single()` would fit better. I think we can also simplify by combining the `Where()` and `Single()` to:

```C#
var relatedOrganisation = organisation.RelatedOrganisations.Single(ro => ro.OrganisationId == relatedOrganisationId);
```

src/NHSD.BuyingCatalogue.Organisations.Api/Controllers/OrganisationsController.cs
```diff
@@ -261,5 +261,31 @@ public async Task<ActionResult> CreateRelatedOrganisationAsync(Guid id, [FromBod
...
+
+            organisation.RelatedOrganisations.Remove(relatedOrganisation);
+
+            await organisationRepository.UpdateAsync(organisation);
```
bjss-chris, line 286: I think we could do something like this here:

```c#
if (organisation.RelatedOrganisations.Remove(relatedOrganisation))
    await organisationRepository.UpdateAsync(organisation);
```
This way we will only update if the related orgs have actually changed and we return `NoContent()` in both cases where we've deleted the relationship and where the relationship didn't exist – I don't think we need to differentiate.

bjss-chris (changes requested):

src/NHSD.BuyingCatalogue.Organisations.Api/Controllers/OrganisationsController.cs
```diff
@@ -261,5 +261,32 @@ public async Task<ActionResult> CreateRelatedOrganisationAsync(Guid id, [FromBod
...
+
+            if (organisation is null)
+            {
+                return NotFound();
+            }
+
+            if (!organisation.RelatedOrganisations.Any(ro => ro.OrganisationId == relatedOrganisationId))
+            {
+                return NoContent();
+            }
```
bjss-chris, line 280: I think we can remove this `if` statement – see comment below.

src/NHSD.BuyingCatalogue.Organisations.Api/Controllers/OrganisationsController.cs
```diff
@@ -261,5 +261,32 @@ public async Task<ActionResult> CreateRelatedOrganisationAsync(Guid id, [FromBod
...
+
+            var relatedOrganisation = organisation.RelatedOrganisations.Single(ro => ro.OrganisationId == relatedOrganisationId);
```
bjss-chris, line 282: If we update this to `SingleOrDefault`, I think we should be able to remove the `if` statement above without requiring any additional changes (`Remove` will accept a null reference).

bjss-chris (changes requested):
Added a suggestion that I think will make the controller method slightly cleaner. Hopefully you agree 😃 but let me know if not.

src/NHSD.BuyingCatalogue.Organisations.Api/Controllers/OrganisationsController.cs
leerogers-bjss: that's true, but i'm not 100% sure how EF will handle it.
should be fine, but gona double check

PR #314, opened 2021-04-26
Update Organisation Repository and Profile Service to Include Related Organisations

leerogers-bjss:
Story: [AB#12958](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/12958)
Tasks:
[AB#13886](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/13886)
[AB#12887](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/12887)

### DIFF
--- a/src/NHSD.BuyingCatalogue.Identity.Api/Services/ProfileService.cs
+++ b/src/NHSD.BuyingCatalogue.Identity.Api/Services/ProfileService.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.IdentityModel.Tokens.Jwt;
+using System.Linq;
 using System.Security.Claims;
 using System.Threading.Tasks;
 using IdentityModel;
@@ -61,26 +62,29 @@ namespace NHSD.BuyingCatalogue.Identity.Api.Services
             if (user is not null)
             {
                 var claims = GetClaimsFromUser(user);
-                var primaryOrganisationName = await GetPrimaryOrganisationName(user.PrimaryOrganisationId);
-                if (primaryOrganisationName is not null)
-                {
-                    claims.Add(primaryOrganisationName);
-                }
+                claims.AddRange(await GetPrimaryOrganisationClaims(user.PrimaryOrganisationId));
 
                 context.IssuedClaims = claims;
             }
         }
 
-        public async Task<Claim> GetPrimaryOrganisationName(Guid primaryOrganisationId)
+        public async Task<IEnumerable<Claim>> GetPrimaryOrganisationClaims(Guid primaryOrganisationId)
         {
-            Claim claim = null;
-            var organisationName = (await organisationRepository.GetByIdAsync(primaryOrganisationId))?.Name;
-            if (!organisationName.IsNullOrEmpty())
+            List<Claim> claims = new List<Claim>();
+
+            var primaryOrganisation = await organisationRepository.GetByIdAsync(primaryOrganisationId);
+
+            if (primaryOrganisation is not null)
             {
-                claim = new Claim(ApplicationClaimTypes.PrimaryOrganisationName, organisationName!);
+                if (!primaryOrganisation.Name.IsNullOrEmpty())
+                {
+                    claims.Add(new Claim(ApplicationClaimTypes.PrimaryOrganisationName, primaryOrganisation.Name));
+                }
+
+                claims.AddRange(primaryOrganisation.RelatedOrganisations.Select(ro => new Claim(ApplicationClaimTypes.RelatedOrganisationId, ro.OrganisationId.ToString())).ToArray());
             }
 
-            return claim;
+            return claims;
         }
 
         public async Task IsActiveAsync(IsActiveContext context)
--- a/src/NHSD.BuyingCatalogue.Identity.Api/Services/ProfileService.cs
+++ b/src/NHSD.BuyingCatalogue.Identity.Api/Services/ProfileService.cs
@@ -74,15 +74,17 @@ namespace NHSD.BuyingCatalogue.Identity.Api.Services
 
             var primaryOrganisation = await organisationRepository.GetByIdAsync(primaryOrganisationId);
 
-            if (primaryOrganisation is not null)
-            {
-                if (!primaryOrganisation.Name.IsNullOrEmpty())
-                {
-                    claims.Add(new Claim(ApplicationClaimTypes.PrimaryOrganisationName, primaryOrganisation.Name));
-                }
+            if (primaryOrganisation is null)
+                return claims;
 
-                claims.AddRange(primaryOrganisation.RelatedOrganisations.Select(ro => new Claim(ApplicationClaimTypes.RelatedOrganisationId, ro.OrganisationId.ToString())).ToArray());
+            if (!primaryOrganisation.Name.IsNullOrEmpty())
+            {
+                claims.Add(new Claim(ApplicationClaimTypes.PrimaryOrganisationName, primaryOrganisation.Name));
             }
+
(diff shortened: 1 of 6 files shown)


src/NHSD.BuyingCatalogue.Identity.Api/Services/ProfileService.cs
```diff
@@ -61,26 +62,29 @@ public async Task GetProfileDataAsync(ProfileDataRequestContext context)
...
         {
-            Claim claim = null;
-            var organisationName = (await organisationRepository.GetByIdAsync(primaryOrganisationId))?.Name;
-            if (!organisationName.IsNullOrEmpty())
+            List<Claim> claims = new List<Claim>();
+
+            var primaryOrganisation = await organisationRepository.GetByIdAsync(primaryOrganisationId);
+
+            if (primaryOrganisation is not null)
             {
```
bjss-chris, line 78: I think we should invert this `if` – I think the method will be slightly more readable.

src/NHSD.BuyingCatalogue.Identity.Api/Services/ProfileService.cs
```diff
@@ -61,26 +62,29 @@ public async Task GetProfileDataAsync(ProfileDataRequestContext context)
...
-                claim = new Claim(ApplicationClaimTypes.PrimaryOrganisationName, organisationName!);
+                if (!primaryOrganisation.Name.IsNullOrEmpty())
+                {
+                    claims.Add(new Claim(ApplicationClaimTypes.PrimaryOrganisationName, primaryOrganisation.Name));
+                }
+
+                claims.AddRange(primaryOrganisation.RelatedOrganisations.Select(ro => new Claim(ApplicationClaimTypes.RelatedOrganisationId, ro.OrganisationId.ToString())).ToArray());
```
bjss-chris, line 84: Do we need the `ToArray` here?

The line is starting to get a little long too so might be worth moving the lambda a new line.

src/NHSD.BuyingCatalogue.Identity.Common/Constants/ApplicationClaimTypes.cs
```diff
@@ -4,6 +4,8 @@ public static class ApplicationClaimTypes
     {
         public const string PrimaryOrganisationId = "primaryOrganisationId";
         public const string PrimaryOrganisationName = "primaryOrganisationName";
+        public const string RelatedOrganisationId = "relatedOrganisationId";
+        public const string RelatedOrganisationName = "relatedOrganisationName";
```
bjss-chris, line 8: I don't think we need the new name claim.

bjss-chris (changes requested):

src/NHSD.BuyingCatalogue.Identity.Api/Services/ProfileService.cs
leerogers-bjss: inverting the if gives you something like this
```c#
        public async Task<IEnumerable<Claim>> GetPrimaryOrganisationClaims(Guid primaryOrganisationId)
        {
            List<Claim> claims = new List<Claim>();

            var primaryOrganisation = await organisationRepository.GetByIdAsync(primaryOrganisationId);

            if (primaryOrganisation is null)
            {
                return claims;
            }

            if (!primaryOrganisation.Name.IsNullOrEmpty())
            {
                claims.Add(new Claim(ApplicationClaimTypes.PrimaryOrganisationName, primaryOrganisation.Name));
            }

            claims.AddRange(primaryOrganisation.RelatedOrganisations.Select(ro => new Claim(ApplicationClaimTypes.RelatedOrganisationId, ro.OrganisationId.ToString())));

            return claims;
        }
```
which imo doesn't look as good and adds extra lines of code

src/NHSD.BuyingCatalogue.Identity.Api/Services/ProfileService.cs
leerogers-bjss: yeah i can remove the too array and list the lambdas vertically

bjss-chris:
Oh, also forgot – do we not need to update `ProfileServiceTests`? I think there are probably integration tests we should update too – probably in the Login User Claims Validation feature. (We can always create a separate task for the integration tests.)

src/NHSD.BuyingCatalogue.Identity.Api/Services/ProfileService.cs
bjss-chris: I prefer that, although I don't slap braces around single line `if` statements either 😃, so I'd personally format it like this:

```c#
public async Task<IEnumerable<Claim>> GetPrimaryOrganisationClaims(Guid primaryOrganisationId)
{
    List<Claim> claims = new List<Claim>();

    var primaryOrganisation = await organisationRepository.GetByIdAsync(primaryOrganisationId);

    if (primaryOrganisation is null)
        return claims;

    if (!primaryOrganisation.Name.IsNullOrEmpty())
        claims.Add(new Claim(ApplicationClaimTypes.PrimaryOrganisationName, primaryOrganisation.Name));

    claims.AddRange(primaryOrganisation.RelatedOrganisations.Select(
        ro => new Claim(ApplicationClaimTypes.RelatedOrganisationId, ro.OrganisationId.ToString())));

    return claims;
}
```
However, I know some people have irrational fears about brace-less `if` statements! Ultimately though I'm not that fussed – leave it as is if you prefer.

leerogers-bjss: its not the brace-less if statement, i have a personal preference for having a single return path (if possible ofc) so you always know that it's coming out of one place. but your call tbh :)

src/NHSD.BuyingCatalogue.Identity.Api/Services/ProfileService.cs
leerogers-bjss: tried to make this line a little neater for you, and frankly can't find one i like. best i got is this:
```c#
                claims
                    .AddRange(
                    primaryOrganisation.RelatedOrganisations.Select(
                    ro => new Claim(
                                ApplicationClaimTypes.RelatedOrganisationId,
                                ro.OrganisationId.ToString())));
```
but i'm not convinced :/

src/NHSD.BuyingCatalogue.Identity.Api/Services/ProfileService.cs
bjss-chris: Ah, the old single exit point approach – that's old school. I'm not convinced it's really relevant these days – at least not in a well-written method.

Let's go with the inverted `if` – not because I want to wind you up but because I think it's probably more in-keeping with style used elsewhere. 😃

src/NHSD.BuyingCatalogue.Identity.Api/Services/ProfileService.cs
leerogers-bjss: actually got something i can live with, just gona resolve this

leerogers-bjss:
> Oh, almost forgot – do we not need to update `ProfileServiceTests`? I think there are probably integration tests we should update too – probably in the Login User Claims Validation feature. (We can always create a separate task for the integration tests.)

i'll take a gander, was nothing specific task wise but i'll see if i can add a few in, once i fix the unit tests on the ordering side

bjss-chris (approved):
I'll approve this PR so the claims are available for the FE to use but working on the basis that you're looking at updating the tests.

PR #317, opened 2021-04-27
Add Unit and Integration tests for the relatedOrganisationId Claim Type

leerogers-bjss:
Story: [AB#12958](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/12958)

### DIFF
--- a/tests/NHSD.BuyingCatalogue.Identity.Api.IntegrationTests/Features/Login User Claims Validation.feature
+++ b/tests/NHSD.BuyingCatalogue.Identity.Api.IntegrationTests/Features/Login User Claims Validation.feature
@@ -8,6 +8,7 @@ Background:
         | Name           | OdsCode |
         | Organisation 1 | Ods 1   |
         | Organisation 2 | Ods 2   |
+        | Organisation 3 | Ods 3   |
     And Users exist
         | Id     | OrganisationName | FirstName | LastName | Email                  | PhoneNumber | Disabled | Password        | OrganisationFunction |
         | 012345 |                  | Penny     | Lane     | PennyLane@email.com    | 01234567890 | false    | S0mePa$$w0rd    | Buyer                |
@@ -66,7 +67,7 @@ Scenario: Get the claims of a buyer user with invalid primary organisation id
 
 Scenario: Get the claims of a buyer user with valid primary organisation id
     Given a user is logged in
-        | Username            | Password     | Scope   |
+        | Username               | Password     | Scope   |
         | SnakePliskin@email.com | S0mePa$$w0rd | profile |
     And the claims contains the following information
         | ClaimType               | ClaimValue             |
@@ -131,3 +132,36 @@ Scenario: Get the claims of an authority user with invalid primary organisation 
         | scope                | Organisation         |
         | organisation         | Manage               |
         | account              | Manage               |
+@5447
+Scenario: Get the relatedOrganisation claim of an Authority user with valid primary organisation id
+    Given Organisation Organisation 2 has a Parent Relationship to Organisation Organisation 1
+    And a user is logged in
+        | Username             | Password        | Scope        |
+        | PostmanPat@email.com | An0therPa$$w0rd | Organisation |
+    Then the claims contain RelatedOrganisationIds of these Organisations
+        | OrganisationName |
+        | Organisation 1   |
+
+@5447
+Scenario: Get the multiple relatedOrganisation claims of an Authority user with valid primary organisation id
+    Given Organisation Organisation 2 has a Parent Relationship to Organisation Organisation 1
+    And Organisation Organisation 2 has a Parent Relationship to Organisation Organisation 3
+    And a user is logged in
+        | Username             | Password        | Scope        |
+        | PostmanPat@email.com | An0therPa$$w0rd | Organisation |
+    Then the claims contain RelatedOrganisationIds of these Organisations
+        | OrganisationName |
+        | Organisation 1   |
+        | Organisation 3   |
+
+@5447
+Scenario: Get the multiple relatedOrganisation claims of a Buyer user with a valid primary organisation id
+    Given Organisation Organisation 1 has a Parent Relationship to Organisation Organisation 2
+    And Organisation Organisation 1 has a Parent Relationship to Organisation Organisation 3
+    And a user is logged in
+        | Username               | Password     | Scope   |
+        | SnakePliskin@email.com | S0mePa$$w0rd | profile |
+    Then the claims contain RelatedOrganisationIds of these Organisations
+        | OrganisationName |
+        | Organisation 2   |
+        | Organisation 3   |
--- a/tests/NHSD.BuyingCatalogue.Identity.Api.IntegrationTests/Features/Login User Claims Validation.feature
+++ b/tests/NHSD.BuyingCatalogue.Identity.Api.IntegrationTests/Features/Login User Claims Validation.feature
@@ -135,13 +135,13 @@ Scenario: Get the claims of an authority user with invalid primary organisation 
 
 @5447
 Scenario: Get the relatedOrganisation claim of a Buyer user with valid primary organisation id
-    Given Organisation Organisation 2 has a Parent Relationship to Organisation Organisation 1
+    Given Organisation Organisation 1 has a Parent Relationship to Organisation Organisation 2
     And a user is logged in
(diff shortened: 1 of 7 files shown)


tests/NHSD.BuyingCatalogue.Identity.Api.IntegrationTests/Features/Login User Claims Validation.feature
```diff
@@ -131,3 +132,36 @@ Scenario: Get the claims of an authority user with invalid primary organisation
         | scope                | Organisation         |
         | organisation         | Manage               |
         | account              | Manage               |
+@5447
```
bjss-chris, line 135: Just need a blank line after the last scenario.

tests/NHSD.BuyingCatalogue.Identity.Api.IntegrationTests/Features/Login User Claims Validation.feature
```diff
@@ -131,3 +132,36 @@ Scenario: Get the claims of an authority user with invalid primary organisation
...
+Scenario: Get the relatedOrganisation claim of an Authority user with valid primary organisation id
+    Given Organisation Organisation 2 has a Parent Relationship to Organisation Organisation 1
+    And a user is logged in
+        | Username             | Password        | Scope        |
+        | PostmanPat@email.com | An0therPa$$w0rd | Organisation |
```
bjss-chris, line 140: I think it would be better to use either Penny or Snake here rather than Pat. While it's not really going to affect the outcome of the test, related organizations are only really relevant to buyers and I doubt the NHSD organization, which is the primary organization of an authority user, will ever have any related organizations.

(We could update the implementation to only add the `relatedOrganisationId` claims when the user is a buyer but I think it would be unnecessary based on the point above.)

However, from a purely semantic/pedantic point-of-view it would be better to use a buyer user here (and in the other tests you've added.)

tests/NHSD.BuyingCatalogue.Identity.Api.IntegrationTests/Steps/OrganisationsSteps.cs
```diff
@@ -1,8 +1,10 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Linq;
```
bjss-chris, line 3: Do we need this `using`?

tests/NHSD.BuyingCatalogue.Identity.Api.UnitTests/Builders/OrganisationBuilder.cs
```diff
@@ -16,9 +20,22 @@ private OrganisationBuilder()
...
+            return toBuild;
         }
 
         internal static OrganisationBuilder Create() => new();
+
+        internal OrganisationBuilder WithRelatedOrganisations(List<Guid> relatedOrganisations)
+        {
+            this.relatedOrganisations = relatedOrganisations
+                                        .Select((value, index) => new { value, Name = string.Format(CultureInfo.CurrentCulture, "HealthTrust {0}", index) })
+                                        .Select(ro => new Organisation { OrganisationId = ro.value, Name = ro.Name });
```
bjss-chris, line 37: We could always just name the parameter `organisations` to avoid hiding the field of the same name. I don't mind this naming style though – much better than the loathsome underscore prefix for fields that some people seem to adore! 😃

One other small thing, I'd only indent the `.Select` lines by one tab stop to keep the style consistent (and probably add a blank line before the `return`.

tests/NHSD.BuyingCatalogue.Identity.Api.UnitTests/Services/ProfileServiceTests.cs
```diff
@@ -189,8 +189,25 @@ public static async Task GetProfileDataAsync_GivenAnApplicationUserExistsWithOrg
...
+            ApplicationClaimTypes.Organisation)]
+
+        [TestCase(
+            "",
+            Subject,
+            ApplicationClaimTypes.PrimaryOrganisationId,
+            ApplicationClaimTypes.PrimaryOrganisationName,
+            ApplicationClaimTypes.OrganisationFunction,
+            ApplicationClaimTypes.Ordering,
+            ApplicationClaimTypes.Organisation)]
```
bjss-chris, line 210: These test cases are a tad unwieldy aren't they? 😃 I'm wonder if it's worth using a test case source... Do you find something like this any more readable (assuming `GetSomeTestCaseData` is better named)?

```c#
private static IEnumerable<ITestCaseData> GetSomeTestCaseData()
{
    yield return new TestCaseData("someone@email.com", new[]
    {
        Subject,
        Email,
        EmailVerified,
        ApplicationClaimTypes.PrimaryOrganisationId,
        ApplicationClaimTypes.PrimaryOrganisationName,
        ApplicationClaimTypes.OrganisationFunction,
        ApplicationClaimTypes.Ordering,
        ApplicationClaimTypes.Organisation,
    });

    yield return new TestCaseData(string.Empty, new[]
    {
        Subject,
        ApplicationClaimTypes.PrimaryOrganisationId,
        ApplicationClaimTypes.PrimaryOrganisationName,
        ApplicationClaimTypes.OrganisationFunction,
        ApplicationClaimTypes.Ordering,
        ApplicationClaimTypes.Organisation,
    });
}

[TestCaseSource(nameof(GetSomeTestCaseData))]
public static async Task GetProfileDataAsync_GivenApplicationUserWithEmail_ReturnExpectedClaimList(
```

tests/NHSD.BuyingCatalogue.Identity.Api.UnitTests/Services/ProfileServiceTests.cs
```diff
@@ -426,6 +458,72 @@ public static async Task GetProfileDataAsync_GivenUserRepositoryDependency_FindB
                 r => r.GetByIdAsync(It.Is<string>(param => expectedUserId.Equals(param, StringComparison.Ordinal))));
         }
 
+        [Test]
+
```
bjss-chris, line 462: I think we can omit the `[Test]` attribute when we have one or more `[TestCase]` attributes.

bjss-chris (changes requested):
Looks fine overall – some minor suggestions though.

tests/NHSD.BuyingCatalogue.Identity.Api.IntegrationTests/Features/Login User Claims Validation.feature
leerogers-bjss: i put in a test for single and multiple claims for authority and a multiple claims for buyer, i'll update the authority multiple to be a buyer and delete the authority single one then

tests/NHSD.BuyingCatalogue.Identity.Api.UnitTests/Builders/OrganisationBuilder.cs
leerogers-bjss: i'd like to keep it just as relatedOrganisations just so when we come back to it in the future we know exactly what it's on about. even though the function does tell you i think it's just more descriptive

tests/NHSD.BuyingCatalogue.Identity.Api.UnitTests/Services/ProfileServiceTests.cs
leerogers-bjss: the only issue is that each test case is unique, so we'd need a testcasedata for each test as such, which looking through would mean we'd need 3 of these, which wouldn't save any space really.

but i'm not against it, just depends which one you would prefer

tests/NHSD.BuyingCatalogue.Identity.Api.UnitTests/Builders/OrganisationBuilder.cs
bjss-chris: OK, that's fine.

tests/NHSD.BuyingCatalogue.Identity.Api.UnitTests/Services/ProfileServiceTests.cs
bjss-chris: Not too fussed really – there's pros and cons to each so feel free to leave as is.

bjss-chris (approved):

PR #371, opened 2021-06-11
Fix cookie content issue

bjss-chris:
Story: [AB#12959](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/12959)
Task: [AB#15230](https://buyingcatalog.visualstudio.com/c5f97979-5b03-4d10-ba8d-871d0526b408/_workitems/edit/15230)

### DIFF
--- a/src/NHSD.BuyingCatalogue.Identity.Api/Controllers/ConsentController.cs
+++ b/src/NHSD.BuyingCatalogue.Identity.Api/Controllers/ConsentController.cs
@@ -53,7 +53,7 @@ namespace NHSD.BuyingCatalogue.Identity.Api.Controllers
             return View("Error");
         }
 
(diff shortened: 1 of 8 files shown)


bjss-chris:
Fixes the application crash (see screenshot). However, I think we'll need an additional change to invalidate the old format cookie. Will look at that on Monday and create a follow-up PR.

image (1)

bjss-chris:
The minor version bump isn't absolutely necessary for this PR (as it's a bug fix a minor revision would suffice). However, due to a previous PR that had a minor version bump that was rolled back the build numbers and latest docker images were becoming confusing.

PhysicsJon (approved):

